PR #18, opened 2018-01-09
Bump CSI spec version to v0.1.0

NickrenREN:
Update CSI spec dependency to point to v0.1.0 tag

cc @saad-ali  @jsafrane  @lpabon


jsafrane:
lgtm, thanks!

PR #19, opened 2018-01-10
QoL improvements for Makefile and .gitignore

davidz627:
Updated Makefile so that it doesn't recompile stale binaries. Added gitignore entries for common items and binaries


saad-ali:
/lgtm

@jsafrane for review and approval

.gitignore
```diff
@@ -1,3 +1,7 @@
+# Compiled binaries and deployment files 
+/bin/
+/deploy/
```
jsafrane: I'd like to ignore csi-attacher binary and other compiled artifacts explicitly in `deploy/`. The directory can have useful files that should not be ignored.

.gitignore
```diff
@@ -12,3 +16,38 @@
...
+.settings/**
+
+# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
+.idea/
+*.iml
+
+# Vscode files
+.vscode
+
+# This is where the result of the go build goes
```
jsafrane: `go build` does not create anything in `*output*/`

.gitignore
```diff
@@ -1,3 +1,7 @@
+# Compiled binaries and deployment files 
+/bin/
+/deploy/
+
 # Binaries for programs and plugins
```
jsafrane: I think you can prune existing .gitignore too, e.g. we don't use glide.

davidz627:
@jsafrane comments addressed. PTAL

PR #20, opened 2018-01-10
Added gcr.io push to Makefile

davidz627:
Added gcr.io push to Makefile for automatic pushing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,21 @@
 
 .PHONY: all csi-attacher clean test
 
-IMAGE_NAME=docker.io/k8scsi/csi-attacher
-IMAGE_VERSION=latest
+GCR_IMAGE_PATH=gcr.io/google-containers/sig-storage/csi/external-attacher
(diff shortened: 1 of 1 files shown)


davidz627:
WIP: registry name is not final.

davidz627:
Paths finalized. PTAL

saad-ali:
> I am not sure we need two docker repositories for one image. Which one is authoritative then? Most people can't push to gcr.io and it's outside of our control. These repos will get out of sync quickly.

It's a requirement for containers that GKE will depend on. We do this for many of the existing Kubernetes containers. The permissions thing is something that is a known issue and is being worked on. Consider this a manual step that Google team will execute until the permission issues are resolved.

jsafrane:
ok, I'll merge it, please rebase.

saad-ali:
Discussed with Jan offline, we will close this PR, and come up with another process for syncing these images to gcr.

PR #33, opened 2018-03-13
PluginCapability update for external-attacher

NickrenREN:
PluginCapability update for external-attacher
/cc @saad-ali @jsafrane

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -103,21 +103,31 @@ func main() {
 			os.Exit(1)
 		}
 
-		// Find out if the driver supports attach/detach.
-		supportsAttach, err := csiConn.SupportsControllerPublish(ctx)
+		supportsService, err := csiConn.SupportsPluginControllerService(ctx)
 		if err != nil {
 			glog.Error(err.Error())
 			os.Exit(1)
 		}
-		if !supportsAttach {
+		if !supportsService {
 			handler = controller.NewTrivialHandler(clientset)
-			glog.V(2).Infof("CSI driver does not support ControllerPublishUnpublish, using trivial handler")
+			glog.V(2).Infof("CSI driver does not support Plugin Controller Service, using trivial handler")
 		} else {
-			pvLister := factory.Core().V1().PersistentVolumes().Lister()
-			nodeLister := factory.Core().V1().Nodes().Lister()
-			vaLister := factory.Storage().V1alpha1().VolumeAttachments().Lister()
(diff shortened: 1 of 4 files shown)


cmd/csi-attacher/main.go
```diff
@@ -109,7 +109,12 @@ func main() {
 			glog.Error(err.Error())
 			os.Exit(1)
 		}
-		if !supportsAttach {
+		supportsService, err := csiConn.SupportsPluginControllerService(ctx)
+		if err != nil {
+			glog.Error(err.Error())
+			os.Exit(1)
+		}
+		if !supportsAttach || !supportsService {
```
jsafrane: I think you should check for `SupportsPluginControllerService` first and only when the driver supports controller service you can invoke controller methods such as `ControllerGetCapabilities` used in `SupportsControllerPublish`.

pkg/connection/connection.go
```diff
@@ -164,6 +168,30 @@ func (c *csiConnection) SupportsControllerPublish(ctx context.Context) (bool, er
 	return false, nil
 }
 
+func (c *csiConnection) SupportsPluginControllerService(ctx context.Context) (bool, error) {
```
jsafrane: please add unit test for this function.

cmd/csi-attacher/main.go
```diff
@@ -103,21 +103,28 @@ func main() {
...
-			vaLister := factory.Storage().V1alpha1().VolumeAttachments().Lister()
-			handler = controller.NewCSIHandler(clientset, attacher, csiConn, pvLister, nodeLister, vaLister)
-			glog.V(2).Infof("CSI driver supports ControllerPublishUnpublish, using real CSI handler")
+			// Find out if the driver supports attach/detach.
+			supportsAttach, err := csiConn.SupportsControllerPublish(ctx)
+			if err != nil {
+				glog.Error(err.Error())
+				os.Exit(1)
+			}
+			if supportsAttach {
```
jsafrane: Add `else { /* fall back to trivial handler */ }` to this `if`.

NickrenREN: oh, sorry, i misunderstand this check, will update right now

NickrenREN: updated, PTAL @jsafrane , thanks

pkg/connection/connection.go
NickrenREN: done

jsafrane:
Thanks for the PR!

saad-ali:
Thanks @NickrenREN!

PR #37, opened 2018-03-14
Fix travis container builder

lpabon:
Also changed the makefile tag form `0.2.0` to `canary`.
We should be in control of which version is tagged as `0.2.0`
and so we can do that by hand.


lpabon:
Let's do that after all of the repos have good canary containers

Issue #46, opened 2018-04-25
Allow running multiple attachers for one driver


jsafrane:
Right now, each CSI driver must run exactly one external attacher and we recommend running it in a StatefulSet.

We should fix the attacher to do the same leader election as provisioner and update documentation and examples everywhere so the attacher runs in the same deployment as  the provisioner.

PR #48, opened 2018-05-04
Add leader election.

jsafrane:
External attacher can now run in Deployments. One attacher per driver is active, others are standby to simplify the implementation.

It must be explicitly enabled - it needs namespace for a config map where the leader election is stored.

Fixes: #46

### DIFF
--- a/cmd/csi-attacher/leader.go
+++ b/cmd/csi-attacher/leader.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


cmd/csi-attacher/leader.go
```diff
@@ -0,0 +1,83 @@
...
+		RenewDeadline: renewDeadline,
+		RetryPeriod:   retryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(stop <-chan struct{}) {
+				glog.V(2).Info("Became leader, starting")
+				close(elected)
+			},
+			OnStoppedLeading: func() {
+				glog.Error("Stopped leading")
+				os.Exit(1)
```
lpabon: This leader election code is new to me. Why quit and kill the program which will restart the container which may look like many errors to users. Is there another way to just go back into waiting?

jsafrane: There is, but the fact that leader lost its leadership means that something really bad happened to the leader (or API server) - it was not able to update an API object for 30 seconds.

Kubernetes controller-manager dies too:

https://github.com/jsafrane/kubernetes/blob/master/cmd/kube-controller-manager/app/controllermanager.go#L211

lpabon: Cool 👍

PR #69, opened 2018-08-27
Add permissions needed by leader election.

jsafrane:


davidz627:
Should this also be modified here then: https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go#L458 ?
This is the "default cluster role" for the csi-external-attacher in Kubernetes. 

In fact, maybe it shouldn't be duplicated in this repository at all, and just use the system cluster role instead a la: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/deploy/kubernetes/stable/setup-cluster.yaml#L47

jsafrane:
If I do that then I give the attacher permissions to read all config maps across the cluster. I don't think we want that. The attacher should read config maps only in the namespace where it runs.

(On related note, I could use Endpoints for leader election instead of ConfigMaps to be on-par with the provisioner, but I don't think it changes anything regarding RBAC)

davidz627:
I think the change has the config maps permissions in a `ClusterRole` which AFAIK gives access cluster-wide. If it is to be namespaced it would have to be a `Role`

jsafrane:
You're absolutely right, the patch is wrong. And I noticed that I already have Role for leader election config maps that I missed. So this PR is not needed at all.

PR #78, opened 2018-10-15
Add CSI Migration logic for attach/detach

davidz627:
This PR adds migration logic for attach/detach.
This logic will get called upon from the code-paths introduced in: https://github.com/kubernetes/kubernetes/pull/69830

/assign @jsafrane @saad-ali 

/cc @lpabon


jsafrane:
The code looks great, however, can you please add a test that creates PV 
 with GCEPersistentDiskVolumeSource  and tests that CSI is called with the right arguments? You'll probably need to change `testAttacherName` to GCE, that's fine.

davidz627:
@jsafrane I have added unit tests for the translation functionality :+1:

davidz627:
@jsafrane PTAL when you have the time

jsafrane:
/approve
?

PR #85, opened 2018-10-23
start attach controller inside OnStartedLeading callback

mlmhl:
This is a simple code clean PR:

Currently we run the leader election in a single goroutine and start the `CSIAttachController` in another goroutine after becoming leader.

However, as `LeaderElectionConfig` has an `OnStartedLeading` callback, we can start the `CSIAttachController` inside this callback, just like what the [k8s in-tree controllers](https://github.com/kubernetes/kubernetes/blob/2f9b65143e44895887f3d91bd019e49a4b3b7658/cmd/kube-controller-manager/app/controllermanager.go#L236) do. This can make the code more concise.

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -21,7 +21,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"os/signal"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


cmd/csi-attacher/main.go
```diff
@@ -170,30 +188,8 @@ func main() {
...
-
-	// run...
-	stopCh := make(chan struct{})
-	factory.Start(stopCh)
-	if csiFactory != nil {
-		csiFactory.Start(stopCh)
-	}
-	go ctrl.Run(threads, stopCh)
-
-	// ...until SIGINT
```
jsafrane: This piece got lost in this PR. IMO it makes some sense to wait until processing of an event is finished and not interrupt it in the middle.

mlmhl: I am a bit confused if this works. The program will exit immediately after `close(stopCh)` finished, as closing of a channel happens before a receive that returns a zero value because the channel is closed, more details can be found [here](https://golang.org/ref/mem#tmp_7). That means all goroutines will exit immediately too.

At this moment the `CSIAttachController` may be processing an event and will still be interrupted in the middle.

If we want to wait until processing of an event is finished, maybe we need to add another channel which will be closed when these two [sync funcs](https://github.com/kubernetes-csi/external-attacher/blob/068527da76a9cfc18c4bae08d1d453c6f89fabbd/pkg/controller/controller.go#L118) finished and the `main` function should wait for this channel before exiting. But this may make the code too complicated

jsafrane:
Yeah, you're right, return from main() stops all threads.

/lgtm

Issue #83, opened 2018-10-19
Detach failure if node is removed from cluster


mape90:
## Problem
If node with volumes managed by CSI plugin is removed from Kubernetes cluster before volumes are detached then detaching will newer succeed. 

function csiDetach (in pkg/controller/csi_handler.go +296 (master)) is failing always because it tries to get nodeID from kubernetes api and as node does not anymore exists it fails.

## Environment
This behavior was produced in Openstack with Cinder-CSI plugin during scale down of worker nodes. As there Openstack cloud provider deleted the node from Kubernetes cluster when it found out that the virtual machine did not existed anymore. 

## Expected Behavior
Volume detach should not automatically fail if node where it was attached isn't anymore part of the kubernetes cluster. It should still try to detach the volume. The nodeID could be passed as empty string and then the actual detaching implementation could then decide if it possible to detach the volume or if it might already be detached. 

In case of Openstack volumes are automatically detached when virtual machine is deleted.

jsafrane:
Thanks for the report. It's hard to judge what's right in this case.

> The nodeID could be passed as empty string

That's very implementation dependent. Different drivers will do different things, usage of empty string is not described in CSI spec. It SHALL be "ID returned by `NodeGetInfo`", but Kubernetes does not have this information, because the node was deleted.

One thing I could perhaps do here is to store the node ID in VolumeAttachment annotation.
@saad-ali @davidz627, any opinions?

mape90:
Thanks for quick response. Yes it is best not to alter the spec, so yes that my proposal is not possible. I trust that you find some elegant way to solve this.

davidz627:
@jsafrane I like the idea of storing the node ID in the `VolumeAttachment`, that information is very "tied" to the attach operation and so it seems like a natural place to put it. Is is possible to even create a field in the object?

jsafrane:
Yesterday our Kubernetes CSI workgroup decided that we'll put Node ID into VolumeAttachment annotation during attach / ControllerPublish. Downside of this approach is that it's not updated when a driver gets updated and starts reporting different NodeID than before. The driver must accept old Node ID in this case.

/assing
(but patches are welcome)

jsafrane:
@davidz627, sorry, I did not see your comment (github outage?)

I could add a field for it, still I have bad feeling about it. If a driver is updated while a volume is attached, we should use the new NodeID. Using an old Node ID from VolumeAttachment annotation or field feels like workaround.

[...]

PR #86, opened 2018-10-26
Store NodeID in annotation

jsafrane:
So the node / CSINodeInfo object can be deleted and detach continutes to work.

Fixes: #83

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -159,26 +159,44 @@ func (h *csiHandler) syncDetach(va *storage.VolumeAttachment) error {
 	return nil
 }
 
-func (h *csiHandler) addVAFinalizer(va *storage.VolumeAttachment) (*storage.VolumeAttachment, error) {
+func (h *csiHandler) prepareVAFinalizer(va *storage.VolumeAttachment) (newVA *storage.VolumeAttachment, modified bool) {
 	finalizerName := GetFinalizerName(h.attacherName)
 	for _, f := range va.Finalizers {
 		if f == finalizerName {
 			// Finalizer is already present
 			glog.V(4).Infof("VA finalizer is already set on %q", va.Name)
-			return va, nil
+			return va, false
 		}
 	}
 
 	// Finalizer is not present, add it
-	glog.V(4).Infof("Adding finalizer to VA %q", va.Name)
 	clone := va.DeepCopy()
 	clone.Finalizers = append(clone.Finalizers, finalizerName)
+	glog.V(4).Infof("VA finalizer added to %q", va.Name)
+	return clone, true
+}
+
+func (h *csiHandler) prepareVANodeID(va *storage.VolumeAttachment, nodeID string) (newVA *storage.VolumeAttachment, modified bool) {
+	if existingID, ok := va.Annotations[vaNodeIDAnnotation]; ok && existingID == nodeID {
+		glog.V(4).Infof("NodeID annotation is already set on %q", va.Name)
+		return va, false
+	}
+	clone := va.DeepCopy()
+	if clone.Annotations == nil {
+		clone.Annotations = map[string]string{}
+	}
+	clone.Annotations[vaNodeIDAnnotation] = nodeID
+	glog.V(4).Infof("NodeID annotation added to %q", va.Name)
+	return clone, true
+}
+
+func (h *csiHandler) saveVA(va *storage.VolumeAttachment) (*storage.VolumeAttachment, error) {
 	// TODO: use patch to save us from VersionError
-	newVA, err := h.client.StorageV1beta1().VolumeAttachments().Update(clone)
+	newVA, err := h.client.StorageV1beta1().VolumeAttachments().Update(va)
 	if err != nil {
 		return va, err
 	}
-	glog.V(4).Infof("VA finalizer added to %q", va.Name)
+	glog.V(4).Infof("VolumeAttachment %q updated with finalizer and/or NodeID annotation", va.Name)
 	return newVA, nil
 }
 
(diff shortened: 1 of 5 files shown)


jsafrane:
@davidz627, PTAL

pkg/controller/csi_handler.go
```diff
@@ -157,26 +157,44 @@ func (h *csiHandler) syncDetach(va *storage.VolumeAttachment) error {
 	return nil
 }
 
-func (h *csiHandler) addVAFinalizer(va *storage.VolumeAttachment) (*storage.VolumeAttachment, error) {
+func (h *csiHandler) prepareVAFinalizer(va *storage.VolumeAttachment) (*storage.VolumeAttachment, bool) {
```
davidz627: nit: name this bool return var or add a comment as description

pkg/controller/csi_handler.go
```diff
@@ -253,14 +271,20 @@ func (h *csiHandler) csiAttach(va *storage.VolumeAttachment) (*storage.VolumeAtt
...
 	if err != nil {
 		return va, nil, err
 	}
 
-	va, err = h.addVAFinalizer(va)
-	if err != nil {
-		return va, nil, fmt.Errorf("could not add VolumeAttachment finalizer: %s", err)
+	originalVA := va
+	va, finalizerAdded := h.prepareVAFinalizer(va)
+	va, nodeIDAdded := h.prepareVANodeID(va, nodeID)
```
davidz627: it is not clear the copy semantics here. In fact I looked through and it seems like *both* `prepare` functions do a deepcopy of the `va`. Since these two functions are both only called together here, wondering if we can wrap both in one `prepareVA` function that does the deepcopy so we are not repeating work.

pkg/controller/csi_handler.go
```diff
@@ -253,14 +271,20 @@ func (h *csiHandler) csiAttach(va *storage.VolumeAttachment) (*storage.VolumeAtt
...
+	if finalizerAdded || nodeIDAdded {
+		va, err = h.saveVA(va)
+		if err != nil {
+			// va modification failed, return the original va that's still on API server
```
davidz627: is this necessarily true, could we have error type where API Server has been modified yet error isi still returned. How bad is returning the incorrect VA here (it might not be that important).

pkg/controller/csi_handler.go
```diff
@@ -458,9 +482,18 @@ func (h *csiHandler) getNodeID(driver string, nodeName string) (string, error) {
...
 	// Check Node annotation.
 	node, err := h.nodeLister.Get(nodeName)
-	if err != nil {
+	if err == nil {
+		return GetNodeIDFromNode(driver, node)
+	}
+
+	// Check VolumeAttachment annotation as the last resort
+	if va == nil {
 		return "", err
```
davidz627: this error will not be 100% accurate, it is just returning the nodeLister error, but in reality it should also contain info about `va` being nil.

pkg/controller/csi_handler.go
```diff
@@ -458,9 +482,18 @@ func (h *csiHandler) getNodeID(driver string, nodeName string) (string, error) {
...
 	}
+	if nodeID, found := va.Annotations[vaNodeIDAnnotation]; found {
+		return nodeID, nil
+	}
 
-	return GetNodeIDFromNode(driver, node)
+	// return nodeLister.Get error
+	return "", err
```
davidz627: same comment as above, maybe annotate with fact VA also not found

pkg/controller/csi_handler_test.go
```diff
@@ -511,27 +517,27 @@ func TestCSIHandler(t *testing.T) {
...
 			initialObjects: []runtime.Object{pvWithFinalizer(), nodeWithoutAnnotations()},
-			addedVA:        va(false, fin),
+			addedVA:        va(false, fin, ann),
 			expectedActions: []core.Action{
-				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, vaWithAttachError(va(false, fin), "node \"node1\" has no NodeID annotation")),
+				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, vaWithAttachError(va(false, fin, ann), "node \"node1\" has no NodeID annotation")),
 			},
 		},
 		{
 			name:           "CSINodeInfo exists without the driver, Node without annotations -> error",
```
davidz627: wonder if this should work now cause we get the nodeid from the VA

pkg/controller/csi_handler_test.go
```diff
@@ -511,27 +517,27 @@ func TestCSIHandler(t *testing.T) {
 		{
 			name:           "Node without annotations -> error",
 			initialObjects: []runtime.Object{pvWithFinalizer(), nodeWithoutAnnotations()},
-			addedVA:        va(false, fin),
+			addedVA:        va(false, fin, ann),
 			expectedActions: []core.Action{
-				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, vaWithAttachError(va(false, fin), "node \"node1\" has no NodeID annotation")),
+				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, vaWithAttachError(va(false, fin, ann), "node \"node1\" has no NodeID annotation")),
```
davidz627: wonder if this should work now cause we get the nodeid from the VA

pkg/controller/csi_handler.go
jsafrane: The code tries to save error to `va.Status`. It's likely fail if the error was conflict, but who knows... I try to keep `va` to the last state acknowledged by API server.

pkg/controller/csi_handler_test.go
jsafrane: That's a good question. Something has removed the annotation from the node. VA annotation is probably the best option here. I'll fix it.

pkg/controller/csi_handler.go
jsafrane: ack

pkg/controller/csi_handler.go
jsafrane: I can do `DeepCopy()` before calling `prepareVAFinalizer`, but then I do `DeepCopy` even when no change is needed. Now I do two DeepCopies when one is needed and zero copies when none is needed.

I'd like to avoid funky bool parameters to `prepareVAFinalizer` or `prepareVANodeID` if they should do the clone by themselves or it has been already cloned by previous method.

pkg/controller/csi_handler.go
jsafrane: `va` is nil only only caller wants so. For Attach, I don't want to use annotation in the VA, because that's odd. For Detach, I want to use the annotation. It's not an error if it's nil.

pkg/controller/csi_handler.go
jsafrane: same answer as above :-)

pkg/controller/csi_handler.go
jsafrane: I added some comments around.

pkg/controller/csi_handler_test.go
jsafrane: Oh, sorry, this is about attach. In order to attach, a node **must** have annotation with NodeID (or CSINodeInfo instance). I think that node without Node ID is fundamentally broken and attach should fail. At least, the driver is not registered to the node.

pkg/controller/csi_handler.go
davidz627: hmm I see what you mean. I am ok with either option, feel free to revert if you think the other way is better.

davidz627:
/lgtm
/hold
for squash and CI fix

pkg/controller/csi_handler.go
jsafrane: I restored double `DeepCopy()`

jsafrane:
squashed, rebased, restored double `DeepCopy()`. @davidz627, PTAL

PR #89, opened 2018-11-14
Update external-attacher to use CSI spec 1.0

msau42:


msau42:
/assign @saad-ali 
cc @vladimirvivien @lpabon

msau42:
/hold

I had to temporarily disable one of the unit tests to remove the csi-test dependency.  Will update again once csi-test has been cut

PR #90, opened 2018-11-15
Temporarily calling CSINodeInfo API directly instead of using a lister

verult:
Mitigation around https://github.com/kubernetes/kubernetes/issues/71052, which is causing CSI e2e test failures in https://github.com/kubernetes/kubernetes/issues/70760.

/assign @msau42 
/cc @davidz627 @saad-ali


msau42:
lgtm besides test failures

pohly:
Is there a bug open for reverting this change and fixing the underlying issue? Or how is it getting tracked?

verult:
@pohly I updated the comment in https://github.com/kubernetes/kubernetes/issues/71052 to track the reverting of this PR. Thanks for pointing that out!

PR #96, opened 2018-11-20
Specify the number of worker threads

### LINKED ISSUE #95: Specify the number of work threads
Specify the number of work threads, currently it is set 10 in the code.

cwdsuzhou:
Use flag "thread" to specify the number of work threads
Fixes #95

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -36,9 +36,6 @@ import (
 )
 
 const (
-	// Number of worker threads
-	threads = 10
-
 	// Default timeout of short CSI calls like GetPluginInfo
 	csiTimeout = time.Second
 
@@ -59,6 +56,7 @@ var (
 	enableLeaderElection    = flag.Bool("leader-election", false, "Enable leader election.")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where this attacher runs.")
 	leaderElectionIdentity  = flag.String("leader-election-identity", "", "Unique idenity of this attcher. Typically name of the pod where the attacher runs.")
+	threads                 = flag.Int("threads", 10, "Number of worker threads.")
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cwdsuzhou:
I signed it

jsafrane:
@cwdsuzhou, thanks for the PR. The code looks ok, however, can you please elaborate **why** is it necessary to have number of threads configurable? Is `10` too few? Too much?

cwdsuzhou:
> @cwdsuzhou, thanks for the PR. The code looks ok, however, can you please elaborate **why** is it necessary to have number of threads configurable? Is `10` too few? Too much?

I think it is not very good to set the number threads in the code and it is more flexible to make it as a parameter. If it shows bad performance in the further, users can change the number without changing the code. Currently, in our scenario,  setting the number `2` seems already enough.

So, I request this PR.

jsafrane:
It's not number of OS threads but nr. of go gorutines. They're cheap and don't block any OS resources.

Currently I don't see any reason why every tiny detail of the attacher should be configurable, it should just work out of the box. In similar fashion, you can't configure nr. of worker threads in any Kubernetes controller.

cwdsuzhou:
> It's not number of OS threads but nr. of go gorutines. They're cheap and don't block any OS resources.
> 
> Currently I don't see any reason why every tiny detail of the attacher should be configurable, it should just work out of the box. In similar fashion, you can't configure nr. of worker threads in any Kubernetes controller.

Maybe you are not right. This no. of default controllers in k8s  also  can be configured. You can see them here.
https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/

image

image

cwdsuzhou:
Maybe, `thread` is not very appropriate.

cwdsuzhou:
@davidz627 @sbezverk  can you review the code

sbezverk:
@cwdsuzhou I agree with @jsafrane this parameter defines a number of go routines started by the controller. I am not sure if it needs to be controller by a parameter.
See this line of code confirming that it is just a number of go routines:
https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/controller/controller.go#L117

PR #101, opened 2018-11-26
Update CSINodeInfo

jsafrane:
Kubernetes 1.13 introduced `CSINodeInfo.Spec` + `Status`. Fix the code and vendor kubernetes 1.13.0-beta2.

Fixes: #99

@xing-yang, I fixed your PR to vendor the right code from the right places (I think :-).

PR #103, opened 2018-11-29
Cherry-pick fixes into 1.0

jsafrane:
Contains:
* https://github.com/kubernetes-csi/external-attacher/pull/98
* https://github.com/kubernetes-csi/external-attacher/pull/101

@saad-ali @msau42 @davidz627 @pohly PTAL


Gopkg.toml
```diff
@@ -20,18 +20,24 @@
   name = "google.golang.org/grpc"
   version = "1.9.2"
 
-[[constraint]]
+[[override]]
+  name = "github.com/json-iterator/go"
+  version = "1.1.4"
+
+// TODO: use [[constraint]] when kubernetes-csi/kubernetes-csi-migration-library uses 1.13 code
```
davidz627: you can pull in the newest version of migration library now

jsafrane: I think it does not make much sense to update for 1.13 beta, I'd wait until 1.13 is released and then update everything.

PR #109, opened 2018-12-14
release tools

pohly:
This improves building and pushing images in various ways:

- rules are shared between CSI repos via https://github.com/kubernetes-csi/csi-release-tools
- automatically derives image tag from repo tags
- never overwrites an existing release image
- no false negative in the "go fmt" check when the number of lines
  contains a 0 digit, prints diff for formatting errors
- updates Go in Travis CI to 1.11.1 (same as in other repos)


build-rules/build.make
```diff
@@ -0,0 +1,103 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The following make rules support building one or more command,
+# packaging them in a containers and pushing to a registry. The expected
+# repository layout is:
+# - cmd/*/*.go - source code for each command
+# - cmd/*/Dockerfile - docker file for each command
```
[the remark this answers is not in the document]
msau42: Actually, do we have any repos where there are multiple binaries being built?  If not, can we keep the Dockerfile at the root?  Moving the Dockerfile will break existing CI and make it more complicated to manage the build process across release branches.

build-rules/travis.yml
```diff
@@ -0,0 +1,14 @@
+language: go
+sudo: required
+services:
+  - docker
+matrix:
+  include:
+  - go: 1.10.3
```
msau42: Can we update this to 1.11? I believe some repos like external-provisioner have already been updated

build-rules/build.make
```diff
@@ -0,0 +1,103 @@
...
+			: "release image $(IMAGE_NAME):$$tag already exists, skipping push"; \
+		fi; \
+	done
+
+build: $(CMDS:%=build-%)
+container: $(CMDS:%=container-%)
+push: $(CMDS:%=push-%)
+
+clean:
+	-rm -rf bin
```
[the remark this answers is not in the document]
pohly: Make won't echo the command during execution.

build-rules/build.make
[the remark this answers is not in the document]
pohly: I don't think we do. "drivers" would be the only candidate, but even that only ever published the hostpath driver.

We can support both: use `cmd/*/Dockerfile` if it exists, otherwise `Dockerfile` in the root.

build-rules/travis.yml
pohly: Sure.

That only external-provisioner was updated shows that we need a mechanism like this that makes it simpler to ensure consistency.

pohly:
I've experimented with `git subtree` and rather like the result. It was very easy to move changes back and forth between a local copy (like in this repo) and the shared build rules repo.

The only downside is that we need yet another new repo. We could put it into `csi-lib-utils`, but then `git subtree` will import also all the other files from there.

Let me propose the following:
- we create a new `kubernetes-csi/csi-build-rules` repo with content similar to https://github.com/pohly/csi-build-rules/
- each repo imports and uses that as shown in this PR

Note that there isn't much that is "CSI" specific about that new repo, only the "quay.io/k8scsi" image registry default. I decided to keep the "csi-" prefix nonetheless because just "build-rules" would be fairly generic. I don't have strong opinions about this, though. This PR needs to be refreshed anyway once the repo exists and has been populated.

pohly:
In the CSI WG meeting on Monday we settled on "csi-release-tools" as name for the new repo and "release-tools" as directory name.

New repo requested here: https://github.com/kubernetes/org/issues/342

humblec:
@pohly I like this new repo with build tools in it and the trick with `git subtree`. I havent experimented with `subtree` much so not sure if there are some caveats though :).

release-tools/README.md
```diff
@@ -0,0 +1,51 @@
...
+still be created for maintenance releases later if needed.
+
+Release branches are expected to be named `release-x.y` for releases
+`x.y.z`. Building from such a branch creates `x.y-canary`
+images. Building from master creates the main `canary` image.
+
+Sharing and updating
+--------------------
+
+[`git subtree`](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt)
```
msau42: I'm not familiar with git subtree, so I have a few questions on the general process:
* Do we need to make a PR in every sidecar repo if we want to sync the subtree to the latest?
* Is there protection against changing these files in the sidecar repos instead of the main one?

pohly: > Do we need to make a PR in every sidecar repo if we want to sync the subtree to the latest?

Yes. In that regard it is the same as vendoring.

> Is there protection against changing these files in the sidecar repos instead of the main one?

No. This is something that needs to be checked during the code review.

Technically, a sidecar repo can have local commits touching the `release-tools` content. Such modifications will even survive a resync with upstream, in contrast to `dep`, which will wipe out all local modifications. But there's a risk of merge conflicts and it would defeat the goal of having shared build rules, so I think we shouldn't allow that.

But it is useful while working on a change:
- try out some change in a sidecar repo
- submit a PR for review and automated testing
- once the PR is green, take the commit from the sidecar repo and publish it as PR against the `csi-release-tools` repo; the third line item in the cheat sheet shows how to do that
- when the upstream PR has been merged, do one of:
  - `git subtree pull` in the sidecar (easy, but leads to a slightly more complex git history)
  - start anew locally, `git subtree pull`, then cherry-pick whatever other local changes were in the PR (if none, then this is also simple; the result will be a clean git history without a trace of having worked on this locally first)

pohly: > > Is there protection against changing these files in the sidecar repos instead of the main one?
>
> No. This is something that needs to be checked during the code review.

Let me turn that into a "yes". I have added a check script, see the new commits in this PR. As intended, those changes now cause the PR to not pass CI testing:
https://travis-ci.org/kubernetes-csi/external-attacher/builds/483410807

Note that this will force us into using the "start anew locally" approach for PRs. I personally would enable the check.

msau42: This sounds good to me

pohly: @msau42 Then let's go for it. Please approve https://github.com/kubernetes-csi/csi-release-tools/pull/3 and I'll update this PR here.

That PR was created with `git subtree push` as explained in the README.md.

PR #111, opened 2018-12-19
Scrub secrets 0.4

jarrpa:


msau42:
Looks like go version used by travis is 1.10.  You can update it to 1.11 here: https://github.com/kubernetes-csi/external-attacher/blob/master/.travis.yml

PR #114, opened 2019-01-06
Remove explicit `nil` error initialization.

humblec:
initialization of nil make go-lint unhappy. Removing it via this PR.

### DIFF
--- a/pkg/connection/connection_test.go
+++ b/pkg/connection/connection_test.go
@@ -119,7 +119,7 @@ func TestGetPluginInfo(t *testing.T) {
 		in := &csi.GetPluginInfoRequest{}
 
 		out := test.output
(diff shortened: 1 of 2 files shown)


humblec:
@jsafrane initialization of `nil` make `go-lint` unhappy. Removing it via this PR.

pohly:
Instead of explaining in a PR comment why this change is useful, better put that into the commit message. Then it'll show up automatically in the PR once you submit it and also will get recorded in the git history.

Having said that, the change is good. Thanks for taking the time to clean this up.

/lgtm

humblec:
@pohly sure, will keep it in mind. Also updated in PR subject, so that it will be there in the commit history.

PR #115, opened 2019-01-06
Add function details in external attacher exported functions.

humblec:

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
+// CSIAttachController is a controller that attaches / detaches CSI volumes using provided Handler interface
 type CSIAttachController struct {
 	client        kubernetes.Interface
 	attacherName  string
@@ -104,6 +105,7 @@ func NewCSIAttachController(client kubernetes.Interface, attacherName string, ha
 	return ctrl
 }
 
+// Run starts CSI attacher and listens on channel events
 func (ctrl *CSIAttachController) Run(workers int, stopCh <-chan struct{}) {
 	defer ctrl.vaQueue.ShutDown()
 	defer ctrl.pvQueue.ShutDown()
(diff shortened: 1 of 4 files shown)


pkg/controller/controller.go
```diff
@@ -104,6 +105,7 @@ func NewCSIAttachController(client kubernetes.Interface, attacherName string, ha
 	return ctrl
 }
 
+// Run start CSI attacher and listens on channel events
```
jsafrane: start -> starts

pkg/controller/controller.go
```diff
@@ -38,6 +38,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
+// CSIAttachController contain required interfaces of handlers, eventrecorder, workqueue and listers.
```
jsafrane: contain -> contains

And IMO it's not interesting what *exactly* it contains (handlers, ...), but what is it good for, so "CSIAttachController is a controller that attaches / detaches CSI volumes using provided Handler interface." Or something else in this direction.

pkg/controller/util.go
```diff
@@ -90,6 +90,7 @@ const (
 	vaNodeIDAnnotation         = "csi.alpha.kubernetes.io/node-id"
 )
 
+// SanitizeDriverName sanitize provided driver name.
```
jsafrane: sanitize -> sanitizes

pkg/controller/controller.go
humblec: Sure, Done.

pkg/controller/controller.go
humblec: Done

pkg/controller/util.go
humblec: Done

humblec:
@jsafrane Thanks! comments are addressed. PTAL .

pkg/controller/util.go
pohly: No offense (I really appreciate your work on cleaning up), but this is the kind of comment one gets when the tooling forces one to add something: the comment doesn't add any information that isn't already in the function name and thus adds no value.

I'm now torn between asking for a better comment (for example, one that explains what "sanitize" means in this context) and just ignoring the comment.

jsafrane:
/lgtm
/approve

Thanks for the PR!

PR #119, opened 2019-01-28
Migrate to k8s.io/klog from glog.

humblec:


Gopkg.toml
```diff
@@ -45,6 +45,10 @@
   name = "k8s.io/csi-api"
   version = "kubernetes-1.13.2"
 
+[[constraint]]
+  name = "k8s.io/klog"
+  revision = "8139d8cb77af419532b33dfa7dd09fbc5f1d344f"
```
pohly: See https://github.com/kubernetes-csi/external-provisioner/pull/219#discussion_r253469203

humblec: @pohly I updated the commit message mentioning why `klog` has been updated as `constraint`.  Regarding `revision` , I was locking it to the same revision kubernetes has today https://github.com/kubernetes/kubernetes/blob/master/Godeps/Godeps.json#L4052 . Please let me know if any thing else has to be done/changed here.

pohly: Conceptually, `Godeps.json` corresponds to `Gopkg.lock`. If the goal is to use the same model as in Kubernetes, then a) don't add a revision to `Gopkg.toml` and b) specify a certain revision that you want to update to with `dep ensure -update k8s.io/klog@<revision>`.

Having said that, I don't think that we absolutely need to use exactly the same klog revision as Kubernetes. Using the latest stable version is fine, and as klog doesn't have releases, that means following the master branch.

pohly: Actually, k8s.io/klog does have releases. We should use those and just let dep pick the latest one.

cmd/csi-attacher/main.go
```diff
@@ -67,30 +67,31 @@ var (
 
 func main() {
 	flag.Set("logtostderr", "true")
+	klog.InitFlags(nil)
```
pohly: Hmm, does `flag.Set("logtostderr", "true")` work when it comes before the definition of the `logtostderr` flag itself in `klog.InitFlags(nil)`?

jsafrane: It doesn't :-(

pohly: No, it doesn't, I tried. `klog.InitFlags(nil) must come first.

jsafrane: It's fixed now.

PR #125, opened 2019-02-22
Check if volumeHandler was set

cwdsuzhou:
Check if volumeHandler was set before call "Attach".  If a PV was created by storgeClasses, volumeHandler must not be empty. However, if it was created by a user, it is possible empty, e.g.

apiVersion: v1
kind: PersistentVolume
metadata:
  name: test
  namespace: "default"
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 5Gi
  csi:
    driver: "XXX"
    volumeHandle: " "

I think, attacher should not call "Attach" in this scenario.

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


jsafrane:
CSI does not describe format of `volume_id` and does not explicitly forbid empty space there. In addition, such validation should be done on PV creation so such invalid PV can't be even created.

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L1460

cwdsuzhou:
Thoug CSI does not describe the format of volume_id and does not forbid empty space, it is meaningless for out-tree csi drivers if volume_id is a space.   

https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L1460

Validation forbids empty volume_id, but it does not forbid spaces, so a pv actually can be created

msau42:
I think it should be up to the plugin to validate that the volume handle is in their supported format, since the csi spec does not define any syntax rules.

PR #127, opened 2019-03-02
Refactor external-attacher to use csi-lib-utils/rpc

MorrisLaw:
This PR makes use of the new `csi-lib-utils/rpc` package.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -136,15 +136,16 @@
   version = "v1.1.5"
 
 [[projects]]
-  digest = "1:b09c9ac14d93f481c768e73a12d4e8527ac25232d593f2ad918ffe462901b654"
+  digest = "1:a0892607b4f5385bb9fb12759facc8fad4e61b8b557384e4a078150c6ba43623"
   name = "github.com/kubernetes-csi/csi-lib-utils"
   packages = [
     "connection",
     "protosanitizer",
+    "rpc",
   ]
   pruneopts = "NUT"
-  revision = "e581edfed23c6ce10243bc0efb2292cf68b83e1d"
-  version = "v0.3.1"
+  revision = "8053f37bf1d11d769c20f9514538c4b3b906e1f7"
+  version = "v0.4.0-rc1"
 
 [[projects]]
   digest = "1:0f47ba38b647bb8e7cddb71c3341134b2c2eaa8cef2af82291b5c9870ee7f572"
@@ -688,6 +689,7 @@
     "github.com/golang/protobuf/proto",
     "github.com/kubernetes-csi/csi-lib-utils/connection",
     "github.com/kubernetes-csi/csi-lib-utils/protosanitizer",
+    "github.com/kubernetes-csi/csi-lib-utils/rpc",
     "github.com/kubernetes-csi/csi-test/driver",
     "google.golang.org/grpc",
     "google.golang.org/grpc/codes",
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -43,7 +43,7 @@
 
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-lib-utils"
-  version = "0.3.1"
+  version = ">=0.4.0-rc1"
 
 [prune]
    non-go = true
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -31,8 +31,12 @@ import (
 	csiinformers "k8s.io/csi-api/pkg/client/informers/externalversions"
 	"k8s.io/klog"
 
-	"github.com/kubernetes-csi/external-attacher/pkg/connection"
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-csi/csi-lib-utils/connection"
+	"github.com/kubernetes-csi/csi-lib-utils/rpc"
+	"github.com/kubernetes-csi/external-attacher/pkg/attacher"
 	"github.com/kubernetes-csi/external-attacher/pkg/controller"
+	"google.golang.org/grpc"
 )
 
 const (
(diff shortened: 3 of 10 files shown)


MorrisLaw:
@andrewsykim

pkg/attacher/connection.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
andrewsykim: Let's rename this file to `attacher.go`

pkg/attacher/connection.go
```diff
@@ -23,27 +23,17 @@ import (
...
+	"github.com/kubernetes-csi/csi-lib-utils/rpc"
+
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"k8s.io/klog"
 )
 
-// CSIConnection is gRPC connection to a remote CSI driver and abstracts all
+// Attacher is a gRPC connection to a remote CSI driver and abstracts all
```
andrewsykim:
```
// Attacher implements attach/detach operations against a remote CSI driver
```

pkg/attacher/connection.go
```diff
@@ -57,43 +47,40 @@ type CSIConnection interface {
 	// should retry.
 	Detach(ctx context.Context, volumeID string, nodeID string, secrets map[string]string) (detached bool, err error)
 
-	// Probe checks that the CSI driver is ready to process requests
-	Probe(singleProbeTimeout time.Duration) error
-
 	// Close the connection
 	Close() error
```
andrewsykim: We can remove `Close()` as well because it can just be called from the `conn` field

pkg/attacher/connection.go
```diff
@@ -57,43 +47,40 @@ type CSIConnection interface {
...
 	}
-	return &csiConnection{
+	return &attacher{
 		conn: conn,
 	}, nil
 }
 
-func (c *csiConnection) GetDriverName(ctx context.Context) (string, error) {
-	return connection.GetDriverName(ctx, c.conn)
+func (c *attacher) GetDriverName(ctx context.Context) (string, error) {
```
andrewsykim: I think we can just remove this method and call `rpc.GetDriverName(ctx, conn)` directly

pkg/attacher/connection.go
```diff
@@ -57,43 +47,40 @@ type CSIConnection interface {
...
+	return rpc.GetDriverName(ctx, c.conn)
 }
 
-func (c *csiConnection) Probe(singleProbeTimeout time.Duration) error {
-	return connection.ProbeForever(c.conn, singleProbeTimeout)
+func (c *attacher) Probe(singleProbeTimeout time.Duration) error {
```
andrewsykim: Same here, we can remove the Probe method and just call `rpc.Probe` directly

pkg/attacher/connection.go
```diff
@@ -147,7 +134,7 @@ func (c *csiConnection) Detach(ctx context.Context, volumeID string, nodeID stri
 	return true, nil
 }
 
-func (c *csiConnection) Close() error {
+func (c *attacher) Close() error {
```
andrewsykim: This can be removed

pkg/attacher/connection_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
andrewsykim: rename to `attacher_test.go`

pkg/attacher/connection.go
```diff
@@ -57,43 +47,40 @@ type CSIConnection interface {
...
 
-func (c *csiConnection) Probe(singleProbeTimeout time.Duration) error {
-	return connection.ProbeForever(c.conn, singleProbeTimeout)
+func (c *attacher) Probe(singleProbeTimeout time.Duration) error {
+	return rpc.ProbeForever(c.conn, singleProbeTimeout)
 }
 
-func (c *csiConnection) SupportsControllerPublish(ctx context.Context) (supportsControllerPublish bool, supportsPublishReadOnly bool, err error) {
-	caps, err := connection.GetControllerCapabilities(ctx, c.conn)
+func (c *attacher) SupportsControllerPublish(ctx context.Context) (supportsControllerPublish bool, supportsPublishReadOnly bool, err error) {
```
andrewsykim: I don't think `SupportsControllerPublish` needs to be part of attacher, let's just move this as an unexported function in main.go?

pkg/attacher/connection.go
```diff
@@ -103,16 +90,16 @@ func (c *csiConnection) SupportsControllerPublish(ctx context.Context) (supports
 	return supportsControllerPublish, supportsPublishReadOnly, nil
 }
 
-func (c *csiConnection) SupportsPluginControllerService(ctx context.Context) (bool, error) {
-	caps, err := connection.GetPluginCapabilities(ctx, c.conn)
+func (c *attacher) SupportsPluginControllerService(ctx context.Context) (bool, error) {
```
andrewsykim: same here, I don't think `SupportsPluginControllerService ` needs to be part of attacher, let's just move this as an unexported function in main.go?

jsafrane:
There is something missing in this PR, I see it removes `Probe` and `GetControllerCapabilities` from `connection.go`, but it does not start using the ones from csi-lib-util.

pkg/attacher/connection.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package connection
```
jsafrane: `connection` was thin layer around gRPC connection to a real driver so it can be mocked out in unit tests. IMO, `attacher` is IMO something else.

andrewsykim:
> There is something missing in this PR, I see it removes Probe and GetControllerCapabilities from connection.go, but it does not start using the ones from csi-lib-util.

It was removed from the interface but it was still implemented by `CSIConnection`. @MorrisLaw is fixing that now so we just call `Probe` and `GetControllerCapabilities directly from `rpc` package.

andrewsykim:
/lgtm 

Thanks @MorrisLaw! 

/assign @msau42 @jsafrane 
for approval

jsafrane:
I still don't like interface named `Attacher`, but I did not find a better name. So...

/approve

PR #129, opened 2019-03-05
Fix #128 - Cannot attach raw block volumes

### LINKED ISSUE #128: Raw block volumes fail to attach
CSI has supported raw block volumes since version 0.1, and Kubernetes since v1.9, but the external-attacher cannot attach a block volume.

When attaching raw block volumes we don't even see an error on the external-attacher logs saying it cannot handle it, it just goes ahead assuming it's a mount type attach.

Akrog:
CSI has supported raw block volumes since version 0.1, and Kubernetes
since v1.9, but the external-attacher cannot attach a block volume.

When attaching a raw block volume we don't even see an error on the
external-attacher logs saying it cannot handle it, it just goes ahead
assuming it's a mount type of volume.

This patch checks the volume mode on the spec to see if it's a raw
volume or not.

The the volume mode is define and it's PVBlock we'll use the Block
access type, in any other case (undefined and Filesystem mode) we'll use
the Mount access type.

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -146,19 +146,30 @@ func GetVolumeCapabilities(pv *v1.PersistentVolume, csiSource *v1.CSIPersistentV
 		return nil, fmt.Errorf("CSI volume source was nil")
 	}
 
-	fsType := csiSource.FSType
-	if len(fsType) == 0 {
-		fsType = defaultFSType
-	}
-
-	cap := &csi.VolumeCapability{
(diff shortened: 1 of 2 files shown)


jsafrane:
/lgtm
/approve

Thanks for the PR!

Akrog:
Thanks for the quick review!

msau42:
Maybe CSI sanity also needs a test case for this to ensure that volume plugins are correctly checking capabilities cc @bswartz

PR #133, opened 2019-03-26
Update vendor to bring in updated CSI translation library

leakingtapan:
This is needed as part of CSI migration. Updated `k8s.io/csi-translation-lib` to kubernetes-1.14.0 since the latest translation library is depending on k8s v1.14 released staging repo `k8s.io/api` for `LabelZoneFailureDomain`.

Not sure why there are some behavior change along with the dependency update, so that I have to update a test case to make test happy.

/cc @ddebroy @davidz627 @msau42


ddebroy:
Looks good but how about removing the various constraints in the gopkg.toml in a manner similar to https://github.com/kubernetes-csi/external-provisioner/commits/master/Gopkg.toml

pkg/controller/csi_handler_test.go
```diff
@@ -624,7 +624,7 @@ func TestCSIHandler(t *testing.T) {
 			},
 			expectedCSICalls: []csiCall{
 				{"attach", "projects/UNSPECIFIED/zones/testZone/disks/testpd", testNodeID,
-					map[string]string{"partition": "0"}, noSecrets, readWrite, success, notDetached, noMetadata, 0},
+					map[string]string{"partition": ""}, noSecrets, readWrite, success, notDetached, noMetadata, 0},
```
msau42: This is due to https://github.com/kubernetes/kubernetes/pull/74932

ddebroy: It might be better to separate out this change into a separate commit in this PR (with the main commit focussing on the vendoring update).

leakingtapan: @msau42 so this is expected then

msau42: Yes this is expected

PR #135, opened 2019-03-29
Use leader election library from csi-lib-utils

andrewsykim:
Also adds support for lease based leader election which will be the preferred resource for leader election. We still need to support ConfigMap based leader election for backwards compatibility.


cmd/csi-attacher/main.go
```diff
@@ -60,6 +61,7 @@ var (
 	timeout           = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for attaching or detaching the volume.")
 
 	enableLeaderElection    = flag.Bool("leader-election", false, "Enable leader election.")
+	leaderElectionType      = flag.String("leader-election-type", "configmap", "the type of leader election, options are 'configmap' (default) or 'lease' (recommended)")
 	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where this attacher runs.")
 	leaderElectionIdentity  = flag.String("leader-election-identity", "", "Unique identity of this attacher. Typically name of the pod where the attacher runs.")
```
andrewsykim: I don't think identity is needed anymore since the lib will use the hostname if it's not set. Should we deprecate the flag here now?

jsafrane: Yes, please deprecate it (print a warning when it's used)

pohly: @andrewsykim There's https://github.com/kubernetes-csi/csi-lib-utils/blob/master/deprecatedflags/deprecatedflags.go which can be used for deprecated flags.

andrewsykim: ah, thank you!

andrewsykim:
/hold 

Waiting for https://github.com/kubernetes-csi/csi-lib-utils/pull/24

pohly:
How does that affect RBAC rules?

msau42:
Good point. RBAC rules need to be updated to add Lease. (we still need to keep the configmap rules for backwards compat), but drivers can decide to further refine them as needed.

pohly:
> (we still need to keep the configmap rules for backwards compat)

In which case does that help? If the current code no longer needs them, then they can be removed.

No-one should be using these updated RBAC rules with an older sidecar release, that's not part of our compatibility guarantees.

msau42:
The current code still has to support the old leaderelection method for backwards compatibility.  So we still need rules for both ways.

README.md
```diff
@@ -42,9 +42,7 @@ Note that the external-attacher does not scale with more replicas. Only one exte
 
 * `--leader-election`: Enables leader election. This is useful when there are multiple replicas of the same external-attacher running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
 
-* `--leader-election-namespace <namespace>`: Namespace where the external-attacher runs and where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI.
-
-* `--leader-election-identity <id>`: Unique identity of this replica of external-attacher container. It must be unique among all running replicas of the external-attacher. Pod name populated from Kubernetes DownwardAPI is recommended.
```
msau42: Move this to the Deprecated arguments section

deploy/kubernetes/rbac.yaml
```diff
@@ -51,7 +51,7 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
-# Attacher must be able to work with config map in current namespace
```
msau42: Also add a comment that a driver only needs to include one of these rules depending on which leader election method it chose

README.md
```diff
@@ -42,9 +42,7 @@ Note that the external-attacher does not scale with more replicas. Only one exte
...
+* `--leader-election-type`: The resource type to use for leader election, options are 'configmaps' (default) or 'leases' (recommended)
```
msau42: Should we deprecate the "configmap" option?

msau42:
Vendor commit msg needs to be updated.

Also, could you add an entry to the Changelog file?

CHANGELOG-1.1.md
```diff
@@ -16,6 +16,8 @@
 
 ## Other notable changes
 
+* Add support for Lease based leader election. Enable this by setting `--leader-election-type=leases` ([#135](https://github.com/kubernetes-csi/external-attacher/pull/135))
```
msau42: This can probably go under "Notable features"

CHANGELOG-1.1.md
```diff
@@ -16,6 +16,8 @@
...
+* --leader-election-identity is now deprecated. Setting this flag will do nothing now. ([#135](https://github.com/kubernetes-csi/external-attacher/pull/135))
```
msau42: This should go under "Deprecations"

deploy/kubernetes/deployment.yaml
```diff
@@ -38,17 +38,11 @@ spec:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
             - "--leader-election"
-            - "--leader-election-namespace=$(MY_NAMESPACE)"
```
msau42: should this example use lease?

CHANGELOG-1.1.md
```diff
@@ -3,6 +3,7 @@
 # Deprecations
 
 * Command line flag `-connection-timeout` is deprecated and has no effect.
+* --leader-election-identity is now deprecated. Setting this flag will have no effect on leader election. ([#135](https://github.com/kubernetes-csi/external-attacher/pull/135))
```
msau42: Can you follow the style of the above?

deploy/kubernetes/deployment.yaml
andrewsykim: Yeah it should. Originally left that out since I didn't want anyone re-deploying this on an existing deployment but I don't think that'll be a problem

CHANGELOG-1.1.md
```diff
@@ -3,6 +3,7 @@
 # Deprecations
 
 * Command line flag `-connection-timeout` is deprecated and has no effect.
+* Command line flag `--leader-election-identity` is deprecated and has no effect.
```
msau42: Can we also deprecate the "configmaps" option?

pohly:
> The current code still has to support the old leaderelection method for backwards compatibility. So we still need rules for both ways.

It would be useful to document that in the RBAC file: someone might want to remove the parts that they don't need because they only use one mode.

cmd/csi-attacher/main.go
```diff
@@ -179,18 +187,26 @@ func main() {
...
 	} else {
-		// Leader election was requested.
-		if leaderElectionNamespace == nil || *leaderElectionNamespace == "" {
-			klog.Error("-leader-election-namespace must not be empty")
+		var le leaderElection
+
+		// Name of config map with leader election lock
+		lockName := "external-attacher-leader-" + csiAttacher
+		if *leaderElectionType == "configmaps" {
+			le = leaderelection.NewLeaderElectionWithConfigMaps(clientset, lockName, run)
```
msau42: Can you also print out a warning message saying that the configmap option is deprecated.

cmd/csi-attacher/main.go
```diff
@@ -60,14 +62,20 @@ var (
 	timeout           = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for attaching or detaching the volume.")
 
 	enableLeaderElection    = flag.Bool("leader-election", false, "Enable leader election.")
-	leaderElectionNamespace = flag.String("leader-election-namespace", "", "Namespace where this attacher runs.")
-	leaderElectionIdentity  = flag.String("leader-election-identity", "", "Unique identity of this attacher. Typically name of the pod where the attacher runs.")
+	leaderElectionType      = flag.String("leader-election-type", "configmap", "the type of leader election, options are 'configmaps' (default) or 'leases' (recommended)")
```
msau42: Also add a note here that configmaps is deprecated.

Gopkg.toml
```diff
@@ -1,8 +1,8 @@
 # List of dependecies for CSI attacher
 
-[[constraint]]
+[[override]]
   name = "github.com/container-storage-interface/spec"
-  version = "1.0.0"
+  version = "=1.0.0"
```
xing-yang: This should be updated to 1.1.0 to be consistent with csi-lib-utils 0.6.1.
"override" should be changed to "constraint".

andrewsykim: Upgrading to 1.1.0 breaks a bunch of tests which I don't want to be fixing in this PR, hence the override

andrewsykim:
```
go test `go list ./... | grep -v 'vendor' `
# github.com/kubernetes-csi/external-attacher/vendor/github.com/kubernetes-csi/csi-test/driver
vendor/github.com/kubernetes-csi/csi-test/driver/mock.go:41:5: cannot use servers.Controller (type *MockControllerServer) as type csi.ControllerServer in field value:
	*MockControllerServer does not implement csi.ControllerServer (missing ControllerExpandVolume method)
vendor/github.com/kubernetes-csi/csi-test/driver/mock.go:42:5: cannot use servers.Node (type *MockNodeServer) as type csi.NodeServer in field value:
	*MockNodeServer does not implement csi.NodeServer (missing NodeExpandVolume method)
?   	github.com/kubernetes-csi/external-attacher/cmd/csi-attacher	[no test files]
FAIL	github.com/kubernetes-csi/external-attacher/pkg/attacher [build failed]
ok  	github.com/kubernetes-csi/external-attacher/pkg/controller	1.451s
make: *** [test-go] Error 2
```

I can fix this in a follow-up PR

cmd/csi-attacher/main.go
andrewsykim: done

Gopkg.toml
pohly: That should go away if you also update csi-test to 2.0.0.

andrewsykim: ah, thanks, updated now

Gopkg.toml
```diff
@@ -1,16 +1,16 @@
 # List of dependecies for CSI attacher
 
-[[constraint]]
+[[override]]
```
msau42: This can be constraint now?

CHANGELOG-1.1.md
```diff
@@ -3,6 +3,8 @@
 ## Deprecations
 
 * Command line flag `-connection-timeout` is deprecated and has no effect.
+* Command line flag `--leader-election-identity` is deprecated and has no effect.
+* Command line argument `configmaps` for flag `--leader-election-type` is deprecated in favor of `--leader-election-type=leases`. The default is still `configmaps` for backwards compatibility reasons but will be switched to `leases` in a future release.
```
msau42: I thought of another way to document this. What do you think of [this](https://github.com/kubernetes-csi/external-provisioner/pull/263/commits/80918f5746898d01e14af19163df158904ee15f1)?

andrewsykim: Looks much better, I'll update :)

README.md
```diff
@@ -57,10 +57,6 @@ Note that the external-attacher does not scale with more replicas. Only one exte
 
 * `--leader-election`: Enables leader election. This is useful when there are multiple replicas of the same external-attacher running for one CSI driver. Only one of them may be active (=leader). A new leader will be re-elected when current leader dies or becomes unresponsive for ~15 seconds.
 
-* `--leader-election-namespace <namespace>`: Namespace where the external-attacher runs and where leader election object will be created. It is recommended that this parameter is populated from Kubernetes DownwardAPI.
```
msau42: We still need to keep the namespace argument right?  Do we have to deprecate this as well?

andrewsykim: oops, yeah we need to keep this here. I'm not sure about deprecation though since you can run external-attacher outside the cluster.

README.md
msau42: Ah sorry one more nit. This should be moved to the "Deprecated arguments" section

andrewsykim: yup, done

PR #144, opened 2019-04-24
fix default leader election type

mlmhl:
**What type of PR is this?**
/kind bug

The valid leader election types are `configmaps` and `leases`, but the default value of flag `leaderElectionType` is `configmap`, which is illegal. This PR changes the default value to `configmaps`.

```release-note
The default leader election type will be `configmaps` if not specified in the command line
```

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -50,6 +50,9 @@ const (
 
 	// Name of CSI plugin for dummy operation
 	dummyAttacherName = "csi/dummy"
+
+	leaderElectionTypeLeases     = "leases"
+	leaderElectionTypeConfigMaps = "configmaps"
 )
 
 // Command line flags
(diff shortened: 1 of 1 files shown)


cmd/csi-attacher/main.go
```diff
@@ -66,7 +66,7 @@ var (
 	retryIntervalMax   = flag.Duration("retry-interval-max", 5*time.Minute, "Maximum retry interval of failed create volume or deletion.")
 
 	enableLeaderElection    = flag.Bool("leader-election", false, "Enable leader election.")
-	leaderElectionType      = flag.String("leader-election-type", "configmap", "the type of leader election, options are 'configmaps' (default) or 'leases' (recommended). The 'configmaps' option is deprecated in favor of 'leases'.")
+	leaderElectionType      = flag.String("leader-election-type", "configmaps", "the type of leader election, options are 'configmaps' (default) or 'leases' (recommended). The 'configmaps' option is deprecated in favor of 'leases'.")
```
msau42: Can we define constants for these? It could help avoid typos in the future

mlmhl: Good suggestion, already updated, PTAL :)

PR #149, opened 2019-05-10
Use patch to replace update

cwdsuzhou:
**What type of PR is this?**

/kind feature

Use patch to replace update

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -30,9 +30,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	core "k8s.io/client-go/testing"
+	"k8s.io/klog"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/controller/csi_handler_test.go
```diff
@@ -259,8 +270,12 @@ func TestCSIHandler(t *testing.T) {
 			addedVA:        va(false /*attached*/, "" /*finalizer*/, nil /* annotations */),
 			expectedActions: []core.Action{
 				// Finalizer is saved first
-				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, va(false /*attached*/, fin, ann)),
-				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, va(true /*attached*/, fin, ann)),
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
```
jsafrane: I wish there was a way to pass only the new VA - the code that compares the expected actions with reality *could* get old VA from reactor, get patch from API call and compute new VA. I am not sure how hard would it be and if it's even feasible. Would you please check this possibility? Don't spend too much time on it, this PR is mostly OK even without it.

cwdsuzhou: Thanks, I review the codes, this may be possible. Since I do not have too much time, I would like to send anther PR to do this later. Do you think it is ok? 

Other two places you advised have changed.

PR #150, opened 2019-05-24
Allow building inside a container

dparalen:
This patch allows one to build external-attacher
from inside a running (Docker) container in order
to be able to isolate the build from developer's
environment.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This patch allows one to isolate the build from developer's environment.

Fixes #
N/A

No

```release-note
Introduced the ability to build in a container
```

### DIFF
--- a/release-tools/build.make
+++ b/release-tools/build.make
@@ -60,9 +60,20 @@ endif
 # Specific packages can be excluded from each of the tests below by setting the *_FILTER_CMD variables
 # to something like "| grep -v 'github.com/kubernetes-csi/project/pkg/foobar'". See usage below.
 
+BUILD_IN_DOCKER?=false
+# this file is shared between the csi-* projects; see release-tools/README.md
+DOCKER_BUILDER_WD=/go/src/github.com/kubernetes-csi/$(notdir $(CURDIR))
+DOCKER_BUILDER_IMG?=golang:$(shell grep go: $(CURDIR)/release-tools/travis.yml | cut -d: -f2,2 | tr -d ' ')
+BUILDER_CMD=CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags -static'
+DOCKER_BUILDER_CMD=docker run --rm -v $(CURDIR):$(DOCKER_BUILDER_WD):rw -w $(DOCKER_BUILDER_WD) $(DOCKER_BUILDER_IMG)
+
 build-%:
 	mkdir -p bin
-	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
+ifeq "$(BUILD_IN_DOCKER)" "true"
+	$(DOCKER_BUILDER_CMD) /bin/sh -c "$(BUILDER_CMD) -o ./bin/$* ./cmd/$*"
+else
+	$(BUILDER_CMD) -o ./bin/$* ./cmd/$*
+endif
 
 container-%: build-%
 	docker build -t $*:latest -f $(shell if [ -e ./cmd/$*/Dockerfile ]; then echo ./cmd/$*/Dockerfile; else echo Dockerfile; fi) --label revision=$(REV) .


release-tools/build.make
```diff
@@ -57,12 +57,30 @@ else
 TESTARGS =
 endif
 
+BUILD_IN_DOCKER?=false
+DOCKER_BUILDER_WD=/go/src/github.com/kubernetes-csi/external-attacher
+DOCKER_BUILDER_IMG=golang
+
+vendor: Gopkg.toml Gopkg.lock
```
pohly: Better don't assume that the vendor directory gets managed by dep. We are about to change that.

Why is this necessary? Can't you make the existing "vendor" directory available inside the Docker container?

release-tools/build.make
```diff
@@ -57,12 +57,30 @@ else
 TESTARGS =
 endif
 
+BUILD_IN_DOCKER?=false
+DOCKER_BUILDER_WD=/go/src/github.com/kubernetes-csi/external-attacher
+DOCKER_BUILDER_IMG=golang
```
pohly: travis.yaml contains the authoritative definition of which Go version is meant to be used for building the source. It would be useful to use that information to pick the right image - not sure whether that is possible, though.

release-tools/build.make
dparalen: I just thought being able to run `dep` in container might be usable for folks that need to perform the whole build process (CI) in e.g Debian stable. I'm cool to drop this part though.

release-tools/build.make
dparalen: Lemme take a look; I guess this is going to need some `$(shell...` magic though. Unsure how reliable `grep` could be here...

dparalen:
@pohly thanks for a fast response! I'm going to check what can be done about the Go version

release-tools/build.make
```diff
@@ -60,9 +60,27 @@ endif
 # Specific packages can be excluded from each of the tests below by setting the *_FILTER_CMD variables
 # to something like "| grep -v 'github.com/kubernetes-csi/project/pkg/foobar'". See usage below.
 
-build-%:
+BUILD_IN_DOCKER?=false
+DOCKER_BUILDER_WD=/go/src/github.com/kubernetes-csi/external-attacher
+DOCKER_BUILDER_IMG?=golang:$(shell grep go: $(CURDIR)/release-tools/travis.yml | cut -d: -f2,2 | tr -d ' ')
+
+vendor: Gopkg.toml Gopkg.lock
```
pohly: The PR says that it is meant to enable *building* the source code in Docker, but here you make it possible to *vendor* under Docker, which is something that should only be necessary when editing or maintaining the source code.

In that case I think that developers won't get around installing the Go toolchain on their development system, for example to have access to "go fmt". "make test" also needs Go.

release-tools/build.make
```diff
@@ -60,9 +60,27 @@ endif
 # Specific packages can be excluded from each of the tests below by setting the *_FILTER_CMD variables
 # to something like "| grep -v 'github.com/kubernetes-csi/project/pkg/foobar'". See usage below.
 
-build-%:
+BUILD_IN_DOCKER?=false
+DOCKER_BUILDER_WD=/go/src/github.com/kubernetes-csi/external-attacher
```
pohly: This file is shared among repos via csi-release-tools (see README.md). One cannot hard-code individual repo names here.

dparalen:
@pohly my comment about the reasoning behind the vendor dir was removed; just a TL;DR I think it's beneficial to run the `dep` tool in a container if one's (CI) environment is e.g Debian stable but I'm OK to remove that handling

release-tools/build.make
dparalen: OK, lemme see what I can do about that...

release-tools/build.make
dparalen: ... I'll remove then, no prob

pohly:
> I think it's beneficial to run the dep tool in a container if one's (CI) environment is e.g Debian stable

Why would a CI environment do the vendoring? That's something that developers need to do when changing the code.

dparalen:
@pohly bad example, you're right. Better: if developer's environment is e.g MacOS with a brewed installation of Go newer than for instance Dep requires... But that's a different story, I'll remove the vendor handling

dparalen:
@pohly I didn't notice the release-tools was a sub-repo before, sorry about that.
I think I might have fixed the shared directory issue you just pointed out.
In case it's OK-ish now, I'd close this PR and create a PR against release-tools then...

release-tools/build.make
```diff
@@ -60,9 +60,18 @@ endif
...
+DOCKER_BUILDER_IMG?=golang:$(shell grep go: $(CURDIR)/release-tools/travis.yml | cut -d: -f2,2 | tr -d ' ')
+
 build-%:
 	mkdir -p bin
-	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
+ifeq "$(BUILD_IN_DOCKER)" "true"
+	docker run --rm -v $(CURDIR):$(DOCKER_BUILDER_WD):rw -w $(DOCKER_BUILDER_WD) $(DOCKER_BUILDER_IMG) /bin/sh -c " \
+	  CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags -static' -o ./bin/$* ./cmd/$*"
+else
+	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags -static' -o ./bin/$* ./cmd/$*
```
pohly: The Go invocation is complex enough that it warrants putting it into a variable (`GO_BUILD` ?) and then  expanding that variable in those two lines.

jsafrane:
I have some doubts how useful is this going to be.

> in order to be able to isolate the build from developer's environment.

That is why we use `vendor/`.

You still need to have the same go version in your dev env. as in the official build to ensure e.g. consistent `gofmt` output. IMO building in container won't help you with this, at least not as it is proposed here.

release-tools/build.make
dparalen: can do, I'd probably add `$(GOARCH)` expansion in that case as well...

dparalen: or you rather mean just the `docker...` call itself?

dparalen:
Hi,

> I have some doubts how useful is this going to be.
> 
> > in order to be able to isolate the build from developer's environment.
> 
> That is why we use `vendor/`.
> 
> You still need to have the same go version in your dev env. as in the official build to ensure e.g. consistent `gofmt` output. IMO building in container won't help you with this, at least not as it is proposed here.

do you mean the patch would be more useful if it also allowed a developer to run static (source) code analysis in a container?

Thanks!
milan

dparalen:
@pohly I've just updated the patch with the Go build command expansion; you're right about it, looks much more tidy now...

pohly:
There doesn't seem to be a consensus that this approach of developing code inside a container is useful. Therefore let me close this PR.

/close

PR #152, opened 2019-06-05
update rbac for secret

### LINKED ISSUE #88: Add rbac rules for accessing secrets
They should be optional as not all volume drivers will need it: kubernetes/kubernetes#69122

hoyho:
**What type of PR is this?**

/kind bug

Attach will be fail when StorageClass specific parameters `csi.storage.k8s.io/controller-publish-secret-name` and `csi.storage.k8s.io/controller-publish-secret-namespace`

Fixes #88 

```release-note
None
```

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -34,6 +34,13 @@ rules:
   - apiGroups: ["storage.k8s.io"]
     resources: ["volumeattachments"]
     verbs: ["get", "list", "watch", "update", "patch"]
+#Secret permission is optional.
+#Enable it if you need value from secret.
+#For example, you have key `csi.storage.k8s.io/controller-publish-secret-name` in StorageClass.parameters
+#see https://kubernetes-csi.github.io/docs/secrets-and-credentials.html
+#  - apiGroups: [""]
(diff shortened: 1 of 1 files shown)


hoyho:
Does the external-attacher need other permission？ Currently it has verbs: ["get", "list"] @jsafrane

deploy/kubernetes/rbac.yaml
```diff
@@ -28,6 +28,9 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
```
msau42: Can we make these optional with a comment?  Not all drivers are going to need it and we want to reduce rbac permissions as much as possible by default.

hoyho: Ok, sounds good. Since there's no secret permission by default, should we also add a note at  https://github.com/kubernetes-csi/docs/blob/master/book/src/secrets-and-credentials.md#secrets-and-credentials to remind people to enable here?

msau42: Yes, that's a great idea!

msau42:
/lgtm
/approve
Thanks!

PR #154, opened 2019-06-08
Add support for attachment of inline volumes migrated to CSI

ddebroy:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR adds support in the external-attacher to handle volume attachments associated with inline volumes that do not have a regular PV associated with it.

Related to https://github.com/kubernetes/kubernetes/pull/77703

```release-note
Support attachment of inline volumes migrated to CSI
```


pkg/controller/csi_handler_test.go
```diff
@@ -888,12 +1049,12 @@ func TestCSIHandler(t *testing.T) {
 		{
 			name:           "detach invalid PV reference-> error",
 			initialObjects: []runtime.Object{pvWithFinalizer(), node()},
-			addedVA:        deleted(vaWithNoPVReference(va(true, fin, ann))),
+			addedVA:        deleted(vaWithNoPVReferenceNorInlineVolumeSpec(va(true, fin, ann))),
```
davidz627: I think we're missing a test where both `PV` and `InlineVolumeSpec` are set. Hate to make you write another test after already writing this massive amount, thanks for making sure this is all well tested!

jsafrane:
/approve
Good work! Especially thanks for keeping unit test coverage high!

pkg/controller/csi_handler_test.go
jsafrane: Good catch, we should have an unit test to cover this case. Otherwise I can't find anything wrong in this PR.

ddebroy:
Thanks for reviewing! Comments addressed.

davidz627:
/lgtm
great work!

PR #155, opened 2019-06-12
Add 1.2 changelog with notable feature of inline volume migration

davidz627:
/kind documentation

/assign @ddebroy

```release-note
NONE
```

### DIFF
--- a/CHANGELOG-1.2.md
+++ b/CHANGELOG-1.2.md
@@ -1,0 +1,12 @@
+# Changelog since v1.1.0
+
+## New Features
+
+- Adds CSI Migration support for Azure Disk/File, fixes for backward compatible AccessModes for GCE PD. ([#156](https://github.com/kubernetes-csi/external-attacher/pull/156), [@davidz627](https://github.com/davidz627))
+- Support attachment of inline volumes migrated to CSI ([#154](https://github.com/kubernetes-csi/external-attacher/pull/154), [@ddebroy](https://github.com/ddebroy))
+- Adds --retry-interval-max and --retry-interval-start to the csi-attacher parameters to allow users to limit the exponential backoff retry time for requests. ([#141](https://github.com/kubernetes-csi/external-attacher/pull/141), [@barp](https://github.com/barp))
+
+
+## Bug Fixes
+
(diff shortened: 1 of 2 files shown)


jsafrane:
The external attacher moved from POST to PATCH and RBAC has been updated. I added release-note text to https://github.com/kubernetes-csi/external-attacher/pull/149 just now, please regenerate.

davidz627:
@jsafrane done

CHANGELOG-1.2.md
```diff
@@ -0,0 +1,10 @@
+# Changelog since v1.1.0
+
+## Deprecations
+
+## Notable Features
+
+* Support attachment of inline volumes migrated to CSI [#153](https://github.com/kubernetes-csi/external-attacher/pull/154)
+* Action required: PATCH is used instead of POST to modify VolumeAttachments on API servers. Please note changes in the external attacher RBAC policy. [#149](https://github.com/kubernetes-csi/external-attacher/pull/149)
```
msau42: I believe this is considered a breaking change as part of our still yet to be finalized upgrade policy, ie you can't upgrade just the container image. In fact, our canary tests are failing maybe because of this? https://k8s-testgrid.appspot.com/sig-storage-csi#canary-on-1-14

How much do we really want this change? If we want to keep it, we may need to bump the major version.

davidz627: yeah it was causing GCE PD tests to fail as well. Better to rip off the bandaid quickly rather than leave it for later?
I already updated the RBAC for PD to include PATCH as well :)

msau42: If we want to keep the change, then we should call this release 2.0 to indicate that you may need to take some other actions to make it work.

jsafrane:
I'm reverting PATCH in #157 and updated release note of the original PR, please regenerate again when it gets merged :-(.

CHANGELOG-1.2.md
```diff
@@ -0,0 +1,10 @@
...
+
+## Other notable changes
```
msau42: There were some other changes that I think are worth including too. We should use the release-note tool starting from commit e157b6b514ace191f46d4c879491cb7d11684659.

msau42: Current instructions are here: https://github.com/kubernetes-csi/docs/issues/142#issuecomment-499954471

msau42:
README should be updated to bump K8s version to 1.15

davidz627:
@msau42 resolved with release tool

PR #157, opened 2019-06-18
Revert "Use patch to replace update"

jsafrane:
This reverts commit f95364717f0fb1028185e0276ba20536e941ddaa.

PATCH usage introduced in #149 needs new RBAC rules and that effectively breaks API. We don't want to bump the image to v2.0 (yet).

ddebroy @davidz627 PTAL, I think I fixed in-line attachment unit tests correctly.

@cwdsuzhou, sorry, your contribution is great, however we'd prefer to be on the safe side. We'll keep your patch around when v2.0 time comes.

/kind api-change
(?)

```release-note
NONE
```

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -29,7 +29,6 @@ import (
 	storage "k8s.io/api/storage/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	storagelisters "k8s.io/client-go/listers/storage/v1beta1"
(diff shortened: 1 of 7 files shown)


pkg/controller/csi_handler.go
```diff
@@ -192,8 +191,9 @@ func (h *csiHandler) prepareVANodeID(va *storage.VolumeAttachment, nodeID string
 	return clone, true
 }
 
-func (h *csiHandler) saveVA(va *storage.VolumeAttachment, patch []byte) (*storage.VolumeAttachment, error) {
-	newVA, err := h.client.StorageV1beta1().VolumeAttachments().Patch(va.Name, types.MergePatchType, patch)
+func (h *csiHandler) saveVA(va *storage.VolumeAttachment) (*storage.VolumeAttachment, error) {
+	// TODO: use patch to save us from VersionError
```
davidz627: can you link an issue (with a link to the original reverted Patch PR) for context

msau42:
BTW, opened https://github.com/kubernetes-csi/external-attacher/issues/158 to keep track of changes we should do when we're ready for a 2.0

pkg/controller/csi_handler.go
jsafrane: added links to #158 and #157

PR #163, opened 2019-07-17
Translate PV so that the pvSpec is also translated for things like AccessModes etc

davidz627:
/kind bug
/kind failing-test

Fixes a bug where the `pvSpec` was not translated and thus didn't pick up the backwards compatibility changes made in the `pvSpec` when getting `AccessModes` in particular

/assign @msau42 @ddebroy 
/cc @leakingtapan

```release-note
Fix issue to actually translate backwards compatible access modes for CSI Migration
```

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -240,12 +240,6 @@ func getCSISource(pv *v1.PersistentVolume) (*v1.CSIPersistentVolumeSource, error
 	}
 	if pv.Spec.CSI != nil {
 		return pv.Spec.CSI, nil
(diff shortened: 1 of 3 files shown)


davidz627:
This resolves migration test: `[sig-storage] PersistentVolumes Default StorageClass pods that use multiple volumes should be reschedulable [Slow]`
https://testgrid.k8s.io/sig-gcp-compute-persistent-disk-csi-driver#Migration%20Kubernetes%20Master%20Driver%20Latest

A new attacher release is not urgent, but is always welcome :)

PR #165, opened 2019-08-06
Fixed ControllerUnpublish error handling

### LINKED ISSUE #164: external-attacher marks volumes as detached after final error
When a CSI driver returns [non-trainsient error](https://github.com/kubernetes-csi/external-attacher/blob/e4b1060293c48dbf0c3bff15fe277a0978f210b6/pkg/attacher/attacher.go#L132), corresponding VolumeAttachment is marked as detached (i.e. attacher's finalizer is removed).

```
I0805 16:05:08.948172       1 connection.go:180] GRPC call: /csi.v1.Controller/ControllerUnpublishVolume
I0805 16:05:08.948182       1 connection.go:181] GRPC request: {"node_id":"i-0eb51d1f90e270d91","volume_id":"vol-0766d9cc3c1966b86"}
I0805 16:05:15.323947       1 connection.go:183] GRPC response: {}
I0805 16:05:15.324406       1 connection.go:184] GRPC error: rpc error: code = Internal desc = Could not detach volume "vol-0766d9cc3c1966b86" from node "i-0eb51d1f90e270d91": could not detach volume "vol-0766d9cc3c1966b86" from node "i-0eb51d1f90e270d91": RequestLimitExceeded: Request limit exceeded.
        status code: 503, request id: 10ceab6c-4a6d-4da5-add2-d46d4fdb652a
I0805 16:05:15.324422       1 csi_handler.go:369] Detached "csi-51ad5c68abe99844c08cfce659c0f8375d8b0255391a6b28b543501577b3dee6" with error rpc error: code = Internal desc = Could not detach volume "vol-0766d9cc3c1966b86" from node "i-0eb51d1f90e270d91": could not detach volume "vol-0766d9cc3c1966b86" from node "i-0
eb51d1f90e270d91": RequestLimitExceeded: Request limit exceeded.
        status code: 503, request id: 10ceab6c-4a6d-4da5-add2-d46d4fdb652a
I0805 16:05:15.324459       1 util.go:70] Marking as detached "csi-51ad5c68abe99844c08cfce659c0f8375d8b0255391a6b28b543501577b3dee6"
```

This is wrong, basically any error from `ControllerUnpublish` means that the volume could be still attacher and the attacher must wait for successful response to confirm the volume is detached.

jsafrane:
Any error from ControllerUnpublish can mean that a volume could be still attached (or being detached).

**What type of PR is this?**
/kind bug

Fixes #164

```release-note
Action required: processing of ControllerUnpublish errors has changed. CSI drivers SHOULD return success (0), when a deleted node or volume implies that the volume is detached from the node. The external attacher treats NotFound error as any other error and it assumes that the volume may still be attached to the node. Please check behavior of your CSI driver and fix it accordingly.
```

cc @kubernetes-csi/csi-misc

### DIFF
--- a/pkg/attacher/attacher.go
+++ b/pkg/attacher/attacher.go
@@ -37,11 +37,8 @@ type Attacher interface {
 	// status.
 	Attach(ctx context.Context, volumeID string, readOnly bool, nodeID string, caps *csi.VolumeCapability, attributes, secrets map[string]string) (metadata map[string]string, detached bool, err error)
 
-	// Detach given volume from given node. Note that "detached" is returned on
-	// error and means that the volume is for sure detached from the node.
-	// "false" means that the volume may or may not be detached and caller
-	// should retry.
-	Detach(ctx context.Context, volumeID string, nodeID string, secrets map[string]string) (detached bool, err error)
+	// Detach given volume from given node.
+	Detach(ctx context.Context, volumeID string, nodeID string, secrets map[string]string) error
 }
 
 type attacher struct {
@@ -79,7 +76,7 @@ func (a *attacher) Attach(ctx context.Context, volumeID string, readOnly bool, n
 	return rsp.PublishContext, false, nil
 }
 
-func (a *attacher) Detach(ctx context.Context, volumeID string, nodeID string, secrets map[string]string) (detached bool, err error) {
+func (a *attacher) Detach(ctx context.Context, volumeID string, nodeID string, secrets map[string]string) error {
 	client := csi.NewControllerClient(a.conn)
 
 	req := csi.ControllerUnpublishVolumeRequest{
@@ -88,11 +85,8 @@ func (a *attacher) Detach(ctx context.Context, volumeID string, nodeID string, s
 		Secrets:  secrets,
 	}
 
-	_, err = client.ControllerUnpublishVolume(ctx, &req)
-	if err != nil {
-		return isFinalError(err), err
-	}
-	return true, nil
+	_, err := client.ControllerUnpublishVolume(ctx, &req)
+	return err
 }
 
 func logGRPC(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -24,7 +24,7 @@ import (
 
 	"github.com/kubernetes-csi/external-attacher/pkg/attacher"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	storage "k8s.io/api/storage/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -266,6 +266,7 @@ func TestCSIHandler(t *testing.T) {
 	var success error
 	var readWrite = false
 	var readOnly = true
+	var ignored = false // the value is irrelevant for given call
 
 	tests := []testCase{
 		//
@@ -737,7 +738,7 @@ func TestCSIHandler(t *testing.T) {
 				core.NewUpdateAction(vaGroupResourceVersion, metav1.NamespaceNone, deleted(va(false /*attached*/, "", ann))),
 			},
 			expectedCSICalls: []csiCall{
-				{"detach", testVolumeHandle, testNodeID, noAttrs, noSecrets, readWrite, success, detached, noMetadata, 0},
+				{"detach", testVolumeHandle, testNodeID, noAttrs, noSecrets, readWrite, success, ignored, noMetadata, 0},
 			},
 		},
 		{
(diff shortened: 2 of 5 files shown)


pkg/attacher/attacher.go
```diff
@@ -88,11 +85,8 @@ func (a *attacher) Detach(ctx context.Context, volumeID string, nodeID string, s
 		Secrets:  secrets,
 	}
 
-	_, err = client.ControllerUnpublishVolume(ctx, &req)
-	if err != nil {
-		return isFinalError(err), err
```
msau42: I'm worried this will break detach for plugins that interpret the CSI spec as such:

CSI spec says sp must return NotFound if volume or node is not found: https://github.com/container-storage-interface/spec/blob/master/spec.md#controllerunpublishvolume-errors.

I guess here we were assuming if NotFound was returned, the detach is actually irrelevant/successful.  Two scenarios to consider:

1) SP automatically handles volume or node not found as detached: no Unpublish call is actually needed to clean up. Should the sp go against the CSI spec and return success?

2) SP needs ControllerUnpublish called to clean up its state. Still a question of if it should return success or NotFound error afterwards

jsafrane: I added code that interprets `NotFound` as success and filled https://github.com/container-storage-interface/spec/issues/373 to clarify it in the spec.

pkg/controller/csi_handler_test.go
```diff
@@ -266,6 +266,7 @@ func TestCSIHandler(t *testing.T) {
 	var success error
 	var readWrite = false
 	var readOnly = true
+	var ignored = false // the vlaue is irrelevant for given call
```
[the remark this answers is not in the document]
jsafrane: Fixed

pkg/attacher/attacher.go
```diff
@@ -88,11 +85,22 @@ func (a *attacher) Detach(ctx context.Context, volumeID string, nodeID string, s
...
+	_, err := client.ControllerUnpublishVolume(ctx, &req)
 	if err != nil {
-		return isFinalError(err), err
+		// Consume NotFound error.
+		st, ok := status.FromError(err)
+		if !ok {
+			// This is not gRPC error.
+			return err
+		}
+		if st.Code() == codes.NotFound {
```
davidz627: what about a case where `SP` temporarily cannot "find" the volume or the disk, but the disk is still attached. This should be `NOT_FOUND` error that could be remedied by retrying - disk is not actually detached yet.

I would argue that the `SP` should return `OK` if it has determined that the volume/node is gone in a way that we can assume the volume is detached. This should be up to the `SP` to decide.

msau42: We're assuming all volume plugins automatically detach volumes when a node is deleted, and I'm not confident that we can assume that.

gnufied: I think it is a reasonable compromise until this behaviour is clarified in spec. I do not know any volume type where attachment outlives lifecycle of node.

msau42: Sorry ignore my last comment, I misread SP as CO. We're on the same page :-)

gnufied: Either way I think we are asking plugin authors to reinterpret spec in a way which was previously not documented. 

Esp. for node deletion case - do we have a solution in the meanwhile?

msau42: I think vsphere may have some interesting behaviors around node failure/deletion handling. cc @codenrhoden @vladimirvivien

gnufied: afaik - in vsphere too, when you delete a node in vcenter, the vmdk files are marked as detached.

jsafrane:
I checked AWS, it most porbably returns `Internal` error code when either node or the volume does not exist. Filed https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/330

jsafrane:
Cinder: it returns `NotFound` when node does not exist: https://github.com/kubernetes/cloud-provider-openstack/blob/1f7b6810a83357b4629ba9eb769f7a27c44f43bf/pkg/csi/cinder/controllerserver.go#L203-L207

jsafrane:
Filed https://github.com/kubernetes/cloud-provider-openstack/issues/718 for OpenStack / Cinder

pkg/attacher/attacher.go
msau42: Based on CSI spec meeting this morning, it sounds like there is agreement to relax the NotFound error code so that plugins can decide if the unpublish needs to be retried.  Based on that, I think we should just retry for any error.

And given the fact that all of the csi plugins we've looked at so far are handling this wrong already, we probably should add an "ACTION REQUIRED" to the release note.

jsafrane:
I removed special handling of `NotFound`, now it's handled as any other error, i.e. volume is assumes as still attached.

jsafrane:
And updated release-note.

msau42:
/lgtm

@davidz627, I and a few others discussed this, and we came to the conclusion that this change will require a major version bump because we're significantly changing behavior that may require drivers to change.

jsafrane:
>  this change will require a major version bump because we're significantly changing behavior that may require drivers to change.

At the same time it fixes pretty serious bug that should be fixed in all supported branches. Volumes are marked as detached after "final error" (e.g. `Internal` due to rate limit on GCE).

I could backport safer version of this patch, marking a volume as detached after `NotFound`, i.e. the same behavior as is in the released supported versions. All the other errors would lead to retry.

PR #175, opened 2019-08-15
make worker-threads configurable

### LINKED ISSUE #172: Make worker count configurable
Currently, external-attacher will create 10 go routines when it start.
In some case,  for instance, create 1000 Pod with 1000 PVC.  Some
CSI drivers may execute `ControllerPublishVolume` slowly. 
10 routines may not suitable for some case, some user may need a bigger amount of sync worker.
I think it's necessary to make the worker count configurable.

hoyho:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

 /kind feature

> /kind flake

Currently,  external-attacher has a hard coded thread count.
However, when user create lot of Pod and PVC currently.
10 thread is not enough to sync VolumeAttachment and PV or it will take 
a bit longer time than expected.
Add a configurable flag so that user can configure a suitable value according to their own cluster scale and situation.

Fixes #172 

```release-note
Adds a new flag `--worker-threads` to control the number of goroutines for processing VolumeAttachments. The default value is 10 workers.
```

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -40,8 +40,6 @@ import (
 )
 
 const (
-	// Number of worker threads
-	threads = 10
 
 	// Default timeout of short CSI calls like GetPluginInfo
 	csiTimeout = time.Second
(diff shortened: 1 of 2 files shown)


cmd/csi-attacher/main.go
```diff
@@ -190,7 +189,7 @@ func main() {
 		if csiFactory != nil {
 			csiFactory.Start(stopCh)
 		}
-		ctrl.Run(threads, stopCh)
+		ctrl.Run(*workerThreads, stopCh)
```
jsafrane: Please check that the value is `> 0`

jsafrane:
And please edit the release note - it's a change that worth having there.

cmd/csi-attacher/main.go
hoyho: Fixed. Use Uint flag and check the value before used

cmd/csi-attacher/main.go
```diff
@@ -92,6 +91,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if *workerThreads == 0 {
+		klog.Error("option -worker-threads should be greater than zero")
```
msau42: should => must

cmd/csi-attacher/main.go
```diff
@@ -52,11 +50,12 @@ const (
...
-	resync      = flag.Duration("resync", 10*time.Minute, "Resync interval of the controller.")
-	csiAddress  = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
-	showVersion = flag.Bool("version", false, "Show version.")
-	timeout     = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for attaching or detaching the volume.")
+	kubeconfig    = flag.String("kubeconfig", "", "Absolute path to the kubeconfig file. Required only when running out of cluster.")
+	resync        = flag.Duration("resync", 10*time.Minute, "Resync interval of the controller.")
+	csiAddress    = flag.String("csi-address", "/run/csi/socket", "Address of the CSI driver socket.")
+	showVersion   = flag.Bool("version", false, "Show version.")
+	timeout       = flag.Duration("timeout", 15*time.Second, "Timeout for waiting for attaching or detaching the volume.")
+	workerThreads = flag.Uint("worker-threads", 10, "Number of attacher worker threads")
```
msau42: Can you also update README.md with the new option?

msau42:
I also updated the wording of the release note. ptal

hoyho:
PTAL @msau42  @jsafrane

PR #177, opened 2019-08-16
2.0 prep: Revert "Revert "Use patch to replace update""

jsafrane:
/kind cleanup

This PR restores PATCH support in the external attacher. In v2.0 we can change the behavior and also the default RBAC rules that the attacher needs.

The first commit is the same as originally applied by @cwdsuzhou in #149. The second commit openly admits that the fake client / informer is imperfect and enables tests commented out by #149.

I tried to implement the fake PATCH correctly, however, applying `[]byte` patch to existing API object looks impossible.

```release-note
Action required: The external-attacher now uses PATCH HTTP method to update API objects. Please update the attacher RBAC policy to allow the attacher to `patch`  VolumeAttachments and PersistentVolumes. See `deploy/kubernetes/rbac.yaml` for an example.
```

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -24,7 +24,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["persistentvolumes"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list", "watch"]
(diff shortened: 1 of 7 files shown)


deploy/kubernetes/rbac.yaml
```diff
@@ -24,7 +24,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["persistentvolumes"]
-    verbs: ["get", "list", "watch", "update"]
+    verbs: ["get", "list", "watch", "update", "patch"]
```
msau42: Do we still need update?

jsafrane: I'd prefer to have it there, just in case :-)
(btw, I'd prefer if there were just generic "read" and "write" permissions)

PR #181, opened 2019-09-10
Add v2.0 changelog

jsafrane:
/kind documentation

Adds changelog for v2.0

```release-note
NONE
```

/assign @msau42 

I added extra note to the top about v2.0.

### DIFF
--- a/CHANGELOG-2.0.md
+++ b/CHANGELOG-2.0.md
@@ -1,0 +1,20 @@
+# Changelog since v1.2.0
+
+The version 2.0 is not compatible with v1.x. See Action Required section and update CSI driver manifests.
+
+## Action Required
+
+- The external-attacher now uses PATCH HTTP method to update API objects. Please update the attacher RBAC policy to allow the attacher to `patch`  VolumeAttachments and PersistentVolumes. See `deploy/kubernetes/rbac.yaml` for an example. ([#177](https://github.com/kubernetes-csi/external-attacher/pull/177), [@jsafrane](https://github.com/jsafrane))
+- The `-connection-timeout`, `-leader-election-type` and `-leader-election-identity` flags, deprecated in v1.2, have been removed. Please update your manifests for the external-attacher. Leader election uses `lease` object now. Rolling update from v1.2.y release may not work, as multiple leaders may be elected during the update (one using config maps and another using `lease` object).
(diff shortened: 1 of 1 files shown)


CHANGELOG-2.0.md
```diff
@@ -0,0 +1,19 @@
+# Changelog since v1.2.0
+
+The version 2.0 is not compatible with v1.x. See Action Required section and update CSI driver manifests.
+
+## Action Required
+
+- The external-attacher now uses PATCH HTTP method to update API objects. Please update the attacher RBAC policy to allow the attacher to `patch`  VolumeAttachments and PersistentVolumes. See `deploy/kubernetes/rbac.yaml` for an example. ([#177](https://github.com/kubernetes-csi/external-attacher/pull/177), [@jsafrane](https://github.com/jsafrane))
+- The `-connection-timeout`, `-leader-election-type` and `-leader-election-identity` flags, deprecated in v1.2, have been removed. Please update your manifests for the external-attacher.
```
msau42: Also mention here that only the lease type is supported, and leader election will not properly work when doing a rolling upgrade from a version that previously used the configmaps type.

CHANGELOG-2.0.md
```diff
@@ -0,0 +1,19 @@
...
+  The `-dummy` flag has been removed. Please update your manifests for the external-attacher. ([#173](https://github.com/kubernetes-csi/external-attacher/pull/173), [@jsafrane](https://github.com/jsafrane))
+- Processing of ControllerUnpublish errors has changed. CSI drivers SHOULD return success (0), when a deleted node or volume implies that the volume is detached from the node. The external attacher treats NotFound error as any other error and it assumes that the volume may still be attached to the node. Please check behavior of your CSI driver and fix it accordingly. ([#165](https://github.com/kubernetes-csi/external-attacher/pull/165), [@jsafrane](https://github.com/jsafrane))
```
msau42: should or must? should means that it's optional, but it sounds like detach will never succeed?

jsafrane: "SHOULD" is defined here: https://www.ietf.org/rfc/rfc2119.txt and CSI refers to it.

jsafrane: I see CSI uses "SHALL" (=MUST) instead of SHOULD, I'll fix this

CHANGELOG-2.0.md
jsafrane: Fixed

CHANGELOG-2.0.md
jsafrane: Fixed

PR #182, opened 2019-09-17
Exit on connection loss

jsafrane:
Exist when a driver closes its socket.

1. The attacher caches driver capabilities, which could (in theory) change
after driver restart.

2. New driver may not come up for a while and the attacher should not act
as leader during this time.

/kind cleanup

Fixes #

```release-note
The external attacher now exits when it looses connection to a CSI driver. This speeds up re-election of a new attacher leader that has connection to the driver.
```


jsafrane:
@pohly @msau42, PTAL

I'd prefer to have this in upcoming 2.0 release (for Kubernetes 1.16).

PR #184, opened 2019-09-28
Reconcile VolumeAttachment with ListVolumes Published Nodes

davidz627:
Main code-paths ready for review.

Tested with one dynamically provisioned volume with the native PD CSI Driver. After detaching the volume on the cloud provider -> external-attacher logs show correct identification of state change and update of VolumeAttachment object -> external-attacher succeeds with immediate reattachment of volume.

WIP:
- [x] Depends on CSI Spec `1.2`.
- [x] Depends on https://github.com/kubernetes/kubernetes/pull/83593
- [x] Only reconcile volumes if plugin supports `ListVolumePublishedNodes` capability
- [x] Handle migration cases for volume ID
- [x] Update logging to include verbosity
- [x] Write some unit tests

/kind feature
/assign @jsafrane @msau42 

```release-note
Reconciles VolumeAttachment attachment status with actual back-end volume attachment state if plugin supports LIST_VOLUMES_PUBLISHED_NODES capability.
```

### DIFF
--- a/pkg/attacher/lister.go
+++ b/pkg/attacher/lister.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package attacher
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+
+	"google.golang.org/grpc"
+)
+
+type CSIVolumeLister struct {
+	conn *grpc.ClientConn
+}
+
+// NewVolumeLister provides a new VolumeLister object.
+func NewVolumeLister(conn *grpc.ClientConn) *CSIVolumeLister {
+	return &CSIVolumeLister{
+		conn: conn,
+	}
+}
+
+func (a *CSIVolumeLister) ListVolumes(ctx context.Context) (map[string]([]string), error) {
+	client := csi.NewControllerClient(a.conn)
+
+	p := map[string][]string{}
+
+	tok := ""
+	for {
+		rsp, err := client.ListVolumes(ctx, &csi.ListVolumesRequest{
+			StartingToken: tok,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to list volumes: %v", err)
+		}
+
+		for _, e := range rsp.Entries {
+			p[e.GetVolume().VolumeId] = e.Status.PublishedNodeIds
+		}
+		tok = rsp.NextToken
+
+		if len(tok) == 0 {
+			break
+		}
+	}
+
+	return p, nil
+}
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"fmt"
+	"time"
 
 	"k8s.io/klog"
 
@@ -51,6 +52,9 @@ type CSIAttachController struct {
 	vaListerSynced cache.InformerSynced
 	pvLister       corelisters.PersistentVolumeLister
 	pvListerSynced cache.InformerSynced
+
+	shouldReconcileVolumeAttachment bool
+	reconcileSync                   time.Duration
 }
 
 // Handler is responsible for handling VolumeAttachment events from informer.
@@ -67,22 +71,26 @@ type Handler interface {
 	SyncNewOrUpdatedVolumeAttachment(va *storage.VolumeAttachment)
 
 	SyncNewOrUpdatedPersistentVolume(pv *v1.PersistentVolume)
+
+	ReconcileVA() error
 }
 
 // NewCSIAttachController returns a new *CSIAttachController
-func NewCSIAttachController(client kubernetes.Interface, attacherName string, handler Handler, volumeAttachmentInformer storageinformers.VolumeAttachmentInformer, pvInformer coreinformers.PersistentVolumeInformer, vaRateLimiter, paRateLimiter workqueue.RateLimiter) *CSIAttachController {
+func NewCSIAttachController(client kubernetes.Interface, attacherName string, handler Handler, volumeAttachmentInformer storageinformers.VolumeAttachmentInformer, pvInformer coreinformers.PersistentVolumeInformer, vaRateLimiter, paRateLimiter workqueue.RateLimiter, shouldReconcileVolumeAttachment bool, reconcileSync time.Duration) *CSIAttachController {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: client.CoreV1().Events(v1.NamespaceAll)})
 	var eventRecorder record.EventRecorder
 	eventRecorder = broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: fmt.Sprintf("csi-attacher %s", attacherName)})
 
 	ctrl := &CSIAttachController{
-		client:        client,
-		attacherName:  attacherName,
-		handler:       handler,
-		eventRecorder: eventRecorder,
-		vaQueue:       workqueue.NewNamedRateLimitingQueue(vaRateLimiter, "csi-attacher-va"),
-		pvQueue:       workqueue.NewNamedRateLimitingQueue(paRateLimiter, "csi-attacher-pv"),
+		client:                          client,
+		attacherName:                    attacherName,
+		handler:                         handler,
+		eventRecorder:                   eventRecorder,
+		vaQueue:                         workqueue.NewNamedRateLimitingQueue(vaRateLimiter, "csi-attacher-va"),
+		pvQueue:                         workqueue.NewNamedRateLimitingQueue(paRateLimiter, "csi-attacher-pv"),
+		shouldReconcileVolumeAttachment: shouldReconcileVolumeAttachment,
+		reconcileSync:                   reconcileSync,
 	}
 
 	volumeAttachmentInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
@@ -120,6 +128,15 @@ func (ctrl *CSIAttachController) Run(workers int, stopCh <-chan struct{}) {
 	for i := 0; i < workers; i++ {
 		go wait.Until(ctrl.syncVA, 0, stopCh)
 		go wait.Until(ctrl.syncPV, 0, stopCh)
+	}
+
+	if ctrl.shouldReconcileVolumeAttachment {
+		go wait.Until(func() {
+			err := ctrl.handler.ReconcileVA()
+			if err != nil {
+				klog.Errorf("Failed to reconcile volume attachments: %v", err)
+			}
+		}, ctrl.reconcileSync, stopCh)
 	}
 
 	<-stopCh
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"k8s.io/klog"
@@ -42,6 +43,16 @@ type AttacherCSITranslator interface {
 	RepairVolumeHandle(pluginName, volumeHandle, nodeID string) (string, error)
 }
 
+// Lister implements list operations against a remote CSI driver.
+type VolumeLister interface {
+	// ListVolumes calls ListVolumes on the driver and returns a map with keys
+	// of VolumeID and values of the list of Node IDs that volume is published
+	// on
+	ListVolumes(ctx context.Context) (map[string][]string, error)
+}
+
+var _ VolumeLister = &attacher.CSIVolumeLister{}
+
 // csiHandler is a handler that calls CSI to attach/detach volume.
 // It adds finalizer to VolumeAttachment instance to make sure they're detached
 // before deletion.
@@ -49,11 +60,14 @@ type csiHandler struct {
 	client                  kubernetes.Interface
 	attacherName            string
 	attacher                attacher.Attacher
+	CSIVolumeLister         VolumeLister
 	pvLister                corelisters.PersistentVolumeLister
 	nodeLister              corelisters.NodeLister
 	csiNodeLister           storagelisters.CSINodeLister
 	vaLister                storagelisters.VolumeAttachmentLister
 	vaQueue, pvQueue        workqueue.RateLimitingInterface
+	forceSync               map[string]bool
+	forceSyncMux            sync.Mutex
 	timeout                 time.Duration
 	supportsPublishReadOnly bool
 	translator              AttacherCSITranslator
@@ -66,6 +80,7 @@ func NewCSIHandler(
 	client kubernetes.Interface,
 	attacherName string,
 	attacher attacher.Attacher,
+	CSIVolumeLister VolumeLister,
 	pvLister corelisters.PersistentVolumeLister,
 	nodeLister corelisters.NodeLister,
 	csiNodeLister storagelisters.CSINodeLister,
@@ -78,6 +93,7 @@ func NewCSIHandler(
 		client:                  client,
 		attacherName:            attacherName,
 		attacher:                attacher,
+		CSIVolumeLister:         CSIVolumeLister,
 		pvLister:                pvLister,
 		nodeLister:              nodeLister,
 		csiNodeLister:           csiNodeLister,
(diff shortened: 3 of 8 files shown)


msau42:
cc @misterikkit
Do we also want to handle the reverse scenario? Detach if it got attached out of band?

davidz627:
> Do we also want to handle the reverse scenario? Detach if it got attached out of band?

I thought we didn't want to do this for data safety reasons (don't want to detach a disk if it may be in use)

msau42:
I think to handle the dangling volumes case we may want to do something like that. Maybe for data safety can we do something like only detach if we're trying to attach it to a different node? cc @gnufied

davidz627:
@msau42 @misterikkit @jsafrane 
This is ready for review. Only reason it is still WIP is because of the dependencies that need to be updated and vendored in. The code is for the actual feature is complete :+1:

davidz627:
@jsafrane @misterikkit @msau42 PTAL this is 100% ready for review

pkg/attacher/lister.go
```diff
@@ -0,0 +1,75 @@
...
+
+// NewVolumeLister provides a new VolumeLister object.
+func NewVolumeLister(conn *grpc.ClientConn) VolumeLister {
+	return &volumeLister{
+		conn: conn,
+	}
+}
+
+func (a *volumeLister) ListVolumes(ctx context.Context) (map[string]([]string), error) {
+	client := csi.NewControllerClient(a.conn)
```
tedyu: Since conn is a field, maybe make client a field as well ?

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	defer cancel()
+	published, err := h.volumeLister.ListVolumes(ctx)
+	if err != nil {
+		klog.Fatalf("Failed to ListVolumes: %v", err)
+	}
+
+	// Loop over all volume attachment objects
+	vas, err := h.vaLister.List(labels.Everything())
+	if err != nil {
+		klog.Fatalf("Failed to list VA")
```
tedyu: Is it possible to return error instead of bailing out ?

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+
+		volumeHandle, _, err := GetVolumeHandle(source)
+		if err != nil {
+			klog.Fatalf("Failed to get volume handle %v", err)
+		}
+		attachedStatus := va.Status.Attached
+
+		volumeHandle, err = h.translator.RepairVolumeHandle(source.Driver, volumeHandle, nodeID)
+		if err != nil {
+			klog.Fatalf("Failed to repair volume handle: %v", err)
```
tedyu: An error here seems to be good

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+		}
+
+		// If ListVolumes Attached Status is different, update.
+		found := false
+		for _, gotNodeID := range published[volumeHandle] {
+			if gotNodeID == nodeID {
+				found = true
```
tedyu: Add a break after this line

pkg/attacher/lister.go
davidz627: not sure I follow that logic.  It's a common pattern to save the connection and create a new client with it when necessary

tedyu: Alright

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	if err != nil {
+		return errors.New("failed to list all VolumeAttachment objects")
+	}
+
+	for _, va := range vas {
+		nodeID, ok := va.Annotations[vaNodeIDAnnotation]
+		if !ok {
+			return errors.New("failed to find node ID in VolumeAttachment annotation")
+		}
+		// TODO(dyzz): Maybe all of these should just be an "error" then continue.
```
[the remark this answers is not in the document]
davidz627: ack. The errors are returned then logged at level `klog.Errorf` at the top level.

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	// Loop over all volume attachment objects
+	vas, err := h.vaLister.List(labels.Everything())
+	if err != nil {
+		return errors.New("failed to list all VolumeAttachment objects")
+	}
+
+	for _, va := range vas {
+		nodeID, ok := va.Annotations[vaNodeIDAnnotation]
+		if !ok {
+			return errors.New("failed to find node ID in VolumeAttachment annotation")
```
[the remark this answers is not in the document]
davidz627: Logged at top level, does this message fit your expectations?
https://github.com/kubernetes-csi/external-attacher/pull/184/files#diff-3c862eb54a8e0e161b534e0c67e5379eR135

pkg/controller/csi_handler.go
```diff
@@ -93,6 +96,77 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	}
+
+	// Loop over all volume attachment objects
+	vas, err := h.vaLister.List(labels.Everything())
+	if err != nil {
+		return errors.New("failed to list all VolumeAttachment objects")
+	}
+
+	for _, va := range vas {
+		nodeID, ok := va.Annotations[vaNodeIDAnnotation]
```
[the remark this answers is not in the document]
davidz627: not sure I understand this comment. If a `VA` has deletion timestamp that means that the deletion has been requested. The status on the VA could either be `detached` or `attached` - this new functionality will only change that status if the actual state from the driver is inconsistent with that observed on the `VA`. I think this is appropriate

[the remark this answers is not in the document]
davidz627: for example if a VA has deletion timestamp and the status is `attached`. But the true volume status is `detached`. This loop will+should change the status to `detached`.

davidz627:
> A volume was just attached by the handler, yet this reconciler is processing slightly older ListVolumes and marks the volume as detached.

There are two directions of race condition to think about:
1. We have old ListVolumes ("actual state" has changed) when we compare that state to VolumeAttachment

Lets say actual state was updated from `x` to `y` but we have gotten an "old" actual state `x` and the VA has state `y`. In this case we would "incorrectly" reconcile `VA` to state `x`. This would then get resolved in the next loop as we would get actual state `y` and reconcile `VA` to `y`. Though this is probably not ideal to wait an entire reconcile cycle and could cause errors.

To fix we can reverse the direction and always get the `VA` first before the actual state. This way we have two possibilities:
1a.  the `VA` changes between get and update and we catch it in the second case below
1b. The `VA` is up to date, we get actual state `x`, some other controller changes actual state to `y` before we update the `VA` then:
1bi. `VA` gets updated before we can change it and we're in case 2.
1bii. We set the "old" `VA` to state `x` and the controller that changed state to `y` can now go and change `VA` state to `y` still.

2. We have an up to date ListVolumes (actual state is accurate) when we compare that state to a old VolumeAttachment
  - This should be easy to catch by checking for `ResourceVersion` and making sure that hasn't changed.

msau42:
If all of the controller paths are calling either Update() or Patch() with resource version, then I think that should handle the races?

The only issue is in case 1bii, this would probably cause the VA update in the attach handler to fail, and we would have to retry the attach operation, but I'm not sure how big of an issue that is since the driver should be idempotent, and I'm not sure how frequent this race will be.

We could potentially reduce the window of a race by checking if an operation is in progress, and skipping that volume for this loop or having some other per-volume locking mechanism.

jsafrane:
What if `ReconcileVA` does not update `VolumeAttachment` that something is wrong, but puts the VA silently into the controller workqueue, marked as "force resync"*? The attacher then calls ControllerPublish / ControllerUnpublish first and updates VA only if it failed using already existing code.

*) `ReconcileVA` needs to mark the VA so VA handler knows it should call ControllerPublish  even when the volume looks attached here:
https://github.com/kubernetes-csi/external-attacher/blob/b0b09820457732ebf7588ad111ffd88ea89812ed/pkg/controller/csi_handler.go#L108-L113
(+similar check in detach?)

There won't be any race + volume attachments get fixed automatically, however, user (or A/D controller) won't get a short blip that volume got detached and back attached.

davidz627:
Thanks for the suggestion Jan. After some careful thought here are the reasons why I think we should continue with the current separate reconciler model:
1) From a conceptual standpoint it is more sound to me to have one controller that's job is to reconcile actual state to the desired state (attacher) and a separate loop that reconciles the known state to the actual state (reconcileVA). This is simple and results in eventual consistency even if there are race conditions. The model of injecting imperative forced re-syncs to the attacher work-queue breaks abstractions and increases complexity within the attacher (now it has to branch on this force resync)
2) I think a "short blip that volume got detached and back attached." is actually desirable from a debugging/logs/user perspective since we can tell that the actual state got out of sync with the `VolumeAttachment` state and we can see the timeline of when that was detected + remedied. This is better than trying to fix it silently IMO. If the fix goes wrong it is harder to figure out at what step things broke down.
  2a) In the case when we have a `VolumeAttachment.Status.Attached=true` but the actual state has the volume detached. If there is a long-running `Attach` error that happens every loop, we would be stuck in a state retrying `Attach` in the background while not updating the state for the user or any other controllers that may be watching this object.

I can definitely be convinced of the force resync model but the above are my major concerns. In summary, it feels like a smell to have a "force" something flag as well as to hide actual state (when we know it) from the user and attempt to silently fix it. Let me know what you think.

jsafrane:
> From a conceptual standpoint it is more sound to me to have one controller that's job is to reconcile actual state to the desired state (attacher) and a separate loop that reconciles the known state to the actual state (reconcileVA). This is simple and results in eventual consistency even if there are race conditions.

Not really. I can agree that `va.Status.Attached` gets eventually fixed even with races. What about `va.Status.AttachmentMetadata`? You don't get them from CSI via `ListVolumes`. The first `Attached = true` must be saved by VA handler, together with the metadata. And I would argue that all of them should be saved by it, what if `AttachmentMetadata` changes? (Not that we have facility in A/D controller / kubelet to use that information, but that's a separate issue).

So, can the `ListVolume loop write only `Attached = false` and ignore ` = true`?
* During attach, if a volume is attached and VA is not yet marked so, we can be sure that the VA is in the attacher workqueue.
* During detach, if a volume is attached and VA is already marked as deched, that means that Detach succeeded and removed its finializer from VA and VA most likely already does not exists (it had no finalizer and set deletion timestamp).

> I think a "short blip that volume got detached and back attached." is actually desirable from a debugging/logs/user perspective 

I agree.

>  the case when we have a VolumeAttachment.Status.Attached=true but the actual state has the volume detached. If there is a long-running Attach error that happens every loop, we would be stuck in a state retrying Attach in the background.

That could be fixed if we wanted...

davidz627:
Ack. I changed the functionality to add volumes back to the workqueue with a forced resync flag. This causes intended behavior for reattaching "accidentally detached" volumes with minimal API updates and no known races. For the detach workflow it's unlikely that the forced resync is necessary as it's very diffiult to get into a state where the `Status.Attached=false` and the finalizer still exists and the volume is still attached. However, I don't think its wrong to cover this case as well anyway.

pkg/attacher/lister.go
```diff
@@ -0,0 +1,75 @@
...
+type VolumeLister interface {
+	// ListVolumes calls ListVolumes on the driver and returns a map with keys
+	// of VolumeID and values of the list of Node IDs that volume is published
+	// on
+	ListVolumes(ctx context.Context) (map[string][]string, error)
+}
+
+type volumeLister struct {
+	conn         *grpc.ClientConn
+	capabilities []csi.ControllerServiceCapability
```
jsafrane: `capabilities` does not seem to be used.

pkg/controller/csi_handler.go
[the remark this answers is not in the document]
jsafrane: ack

pkg/controller/csi_handler.go
```diff
@@ -93,6 +101,99 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
 	h.pvQueue = pvQueue
 }
 
+// ReconcileVA lists volumes from the CSI Driver and reconciles the attachment
+// status with the corresponding VolumeAttachment object. If the attachment
+// status of the volume is different from the state on the VolumeAttachment the
+// VolumeAttachment object is patched to the correct state.
+func (h *csiHandler) ReconcileVA() error {
+	ctx, cancel := context.WithTimeout(context.Background(), h.timeout)
```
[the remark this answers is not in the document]
misterikkit: I see the `Run` func takes a stop channel rather than a context. That is worth fixing before or at the same time as plumbing context into this function.

pkg/controller/csi_handler.go
```diff
@@ -93,6 +101,99 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	vas, err := h.vaLister.List(labels.Everything())
+	if err != nil {
+		return errors.New("failed to list all VolumeAttachment objects")
+	}
+
+	// List volumes after getting volume attachment objects to prevent race
+	// condition where actual state is changed between ListVolumes and
+	// VolumeAttachment List. When listing VolumeAttachments first we can
+	// guarantee any race conditions will result in API object UPDATE failures
+	// instead of stomping on the VolumeAttachment with outdated data
```
misterikkit: Is this just because both lists are RPC calls? Calling these in the other order means your actual state might be stale and get written to VA objects, while this order means your VA objects could be stale causing the UPDATE failures.

Are you sure that actual state can't become stale after you call `ListVolumes`? Or is it just a negligible problem at this point? It seems like there are zero RPC calls in the loop below, but the item could sit on the queue for some time.

misterikkit: Also, do we requeue on UPDATE failures?

pkg/controller/csi_handler.go
```diff
@@ -114,8 +215,8 @@ func (h *csiHandler) SyncNewOrUpdatedVolumeAttachment(va *storage.VolumeAttachme
 }
 
 func (h *csiHandler) syncAttach(va *storage.VolumeAttachment) error {
-	if va.Status.Attached {
-		// Volume is attached, there is nothing to be done.
+	if !h.consumeForceSync(va.Name) && va.Status.Attached {
```
misterikkit: Can we be sure that the va we pulled from the queue is the one we pushed in `ReconcileVA`? What if some other operation is sitting near the front of the queue when we reconcile? Hopefully that's fine?

- if the two queued vas agree on state, then the second one will be skipped but that's fine
- if the two queued vas disagree on state, then they will both pass the already attached/detached check?

pkg/controller/csi_handler.go
```diff
@@ -93,6 +101,99 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+	// VolumeAttachment List. When listing VolumeAttachments first we can
+	// guarantee any race conditions will result in API object UPDATE failures
+	// instead of stomping on the VolumeAttachment with outdated data
+	published, err := h.volumeLister.ListVolumes(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to ListVolumes: %v", err)
+	}
+
+	for _, va := range vas {
+		nodeID, ok := va.Annotations[vaNodeIDAnnotation]
```
misterikkit: When does nodeID move from annotation to spec?

pkg/attacher/lister.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"google.golang.org/grpc"
+)
+
+// Lister implements list operations against a remote CSI driver.
+type VolumeLister interface {
+	// ListVolumes calls ListVolumes on the driver and returns a map with keys
+	// of VolumeID and values of the list of Node IDs that volume is published
+	// on
+	ListVolumes(ctx context.Context) (map[string][]string, error)
```
misterikkit: optional: if order never matters, maybe this could return `map[string]sets.String`?

feel free to ignore if you don't want to add a dep on `sets`

pkg/controller/csi_handler.go
```diff
@@ -93,6 +101,99 @@ func (h *csiHandler) Init(vaQueue workqueue.RateLimitingInterface, pvQueue workq
...
+		}
+		attachedStatus := va.Status.Attached
+
+		volumeHandle, err = h.translator.RepairVolumeHandle(source.Driver, volumeHandle, nodeID)
+		if err != nil {
+			klog.Warningf("failed to repair volume handle %s for driver %s: %v", volumeHandle, source.Driver, err)
+			continue
+		}
+
+		// If ListVolumes Attached Status is different, update.
```
misterikkit: Maybe update this comment to, "Check whether the volume is published to this node"

pkg/controller/csi_handler.go
misterikkit: Looks like the workqueue util de-dupes items.
https://github.com/kubernetes/client-go/blob/a32a6f7a3052374fff2c57c5b4b52e64dcd55eed/util/workqueue/doc.go#L21

misterikkit: Also looks like this was discussed extensively above. Sorry for the repeat.

pkg/controller/controller.go
```diff
@@ -67,22 +71,26 @@ type Handler interface {
 	SyncNewOrUpdatedVolumeAttachment(va *storage.VolumeAttachment)
 
 	SyncNewOrUpdatedPersistentVolume(pv *v1.PersistentVolume)
+
+	ReconcileVA() error
```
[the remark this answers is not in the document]
davidz627: I think `main` consumes the handler. But I'm not looking to make too disruptive of a refactor here :) VA is standard

pkg/controller/framework_test.go
```diff
@@ -494,3 +515,17 @@ func (f *fakeCSIConnection) Close() error {
 func (f *fakeCSIConnection) Probe(timeout time.Duration) error {
 	return nil
 }
+
+// TODO: Remove hardcoding for GCE tests and make more general
+type fakeInTreeToCSITranslator struct{}
```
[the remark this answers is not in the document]
davidz627: It translates `GCE` to `CSI` for some funcs and the identity func for some to confirm basic translation is working. I'm not sure why it's included as a part of this PR but it seems necessary to make the handler work

pkg/controller/csi_handler.go
```diff
@@ -243,14 +344,41 @@ func (h *csiHandler) hasVAFinalizer(va *storage.VolumeAttachment) bool {
...
+func (h *csiHandler) getPVSpec(va *storage.VolumeAttachment) (*v1.PersistentVolumeSpec, error) {
+	if va.Spec.Source.PersistentVolumeName != nil {
+		if va.Spec.Source.InlineVolumeSpec != nil {
+			return nil, errors.New("both InlineCSIVolumeSource and PersistentVolumeName specified in VA source")
+		}
+		pv, err := h.pvLister.Get(*va.Spec.Source.PersistentVolumeName)
+		if err != nil {
+			return nil, err
+		}
+		if h.translator.IsPVMigratable(pv) {
```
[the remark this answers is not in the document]
davidz627: It is generally assumed that everything needs to be a CSI object in the external attacher in general. But I've clarified the naming of the function to "getProcessedPVSpec" to imply that something may have been done to the PV in the func.

pkg/controller/controller.go
```diff
@@ -122,9 +130,25 @@ func (ctrl *CSIAttachController) Run(workers int, stopCh <-chan struct{}) {
...
+		}, ctrl.reconcileSync, stopCh)
+	}
+
 	<-stopCh
 }
 
+// reconcileVA calls ListVolumes on the driver controller and reconciles Attach
+// status of all known VA's with the published_node_ids
+func (ctrl *CSIAttachController) reconcileVA() error {
+	klog.V(4).Infof("Reconciling VolumeAttachments with driver backend state at %v", time.Now())
```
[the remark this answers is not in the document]
davidz627: yes :)

pkg/controller/controller.go
```diff
@@ -122,9 +130,25 @@ func (ctrl *CSIAttachController) Run(workers int, stopCh <-chan struct{}) {
...
+			}
+		}, ctrl.reconcileSync, stopCh)
+	}
+
 	<-stopCh
 }
 
+// reconcileVA calls ListVolumes on the driver controller and reconciles Attach
+// status of all known VA's with the published_node_ids
+func (ctrl *CSIAttachController) reconcileVA() error {
```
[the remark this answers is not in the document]
davidz627: I thought I was following some established pattern but now that I look at it that is not the case.

pkg/controller/csi_handler.go
davidz627: this comment is actually out of date so I've removed it. Yeah the concurrency was discussed at length. Plopping the request to forceResync in a shared queue resolves all known race conditions

pkg/controller/csi_handler.go
davidz627: the `forceSync` flag is based on `vaName` so will not affect any other items on the queue. When the queue eventually pops the item we have requeued for force sync it will go into this modified logic

pkg/controller/csi_handler.go
davidz627: `VA.Spec.NodeName` is different from `NodeID` I think. NodeName refers to the Kubernetes node name - node ID is the reference for that node given by the CSI Driver.... I think :) In PD case the `nodeID` is fully qualified with project and zone, but the node name is not.

@msau42 in case I'm misrepresenting this

pkg/attacher/lister.go
davidz627: I remember reading that it is more efficient in golang to iterate over "small lists" to find an item than look it up in a map. I don't expect the list of nodes that a volume is attached to to ever get large (but this can always be fixed if that assumption changes in the future)

pkg/controller/csi_handler.go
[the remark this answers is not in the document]
davidz627: @misterikkit could you create an issue on this repo for that?

pkg/controller/controller.go
```diff
@@ -122,9 +130,25 @@ func (ctrl *CSIAttachController) Run(workers int, stopCh <-chan struct{}) {
 		go wait.Until(ctrl.syncPV, 0, stopCh)
 	}
 
+	if ctrl.shouldReconcileVolumeAttachment {
+		go wait.Until(func() {
```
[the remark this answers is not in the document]
davidz627: xref comment: https://github.com/kubernetes-csi/external-attacher/pull/184#discussion_r359614277

pkg/controller/csi_handler.go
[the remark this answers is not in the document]
misterikkit: https://github.com/kubernetes-csi/external-attacher/issues/198

pkg/attacher/lister.go
misterikkit: fair, I was thinking primarily about readability, but a small loop isn't that bad.

PR #189, opened 2019-10-24
Yank CSI Spec into the future (bump to v1.2.0)

davidz627:
/kind cleanup

/assign @msau42 

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/kubernetes-csi/external-attacher
 go 1.12
 
 require (
-	github.com/container-storage-interface/spec v1.1.0
+	github.com/container-storage-interface/spec v1.2.0
 	github.com/davecgh/go-spew v1.1.1
 	github.com/evanphx/json-patch v4.2.0+incompatible
 	github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef // indirect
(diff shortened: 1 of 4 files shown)


davidz627:
What is the unit test failure here. I can't figure out what it wants from me

msau42:
Have you run "go mod tidy"?  

The instructions under https://github.com/kubernetes-csi/external-attacher/tree/master/release-tools#dependencies-and-vendoring can probably be more clear on the sequence of steps needed to add/update vendor cc  @pohly

davidz627:
> Have you run "go mod tidy"?
> 
> The instructions under https://github.com/kubernetes-csi/external-attacher/tree/master/release-tools#dependencies-and-vendoring can probably be more clear on the sequence of steps needed to add/update vendor cc @pohly

I ran all these steps - `go mod tidy` and `go mod vendor`. Unit tests still don't seem very happy with me.

pohly:
> I ran all these steps - go mod tidy and go mod vendor. Unit tests still don't seem very happy with me.

The unit tests run with Go 1.12.4. Which version did you use? I can imagine that the result of vendoring varies depending on the Go version. If so, the situation is the same as with "go fmt": to be sure that the result is as expected, one has to use the Go version that the project is currently using.

pohly:
> I can imagine that the result of vendoring varies depending on the Go version.

Note that I haven't actually seen that myself - if this turns out to be indeed the problem, then of course the instructions need to be updated to mention this version dependency.

davidz627:
@pohly I'm  using go 1.13. Can i suggest since unit tests do a strict check against some expected state (so we obviously know how to generate what is expected) that a tool is provided to generate that state? I suggest it could be in a container to avoid all these external tool versioning issues. This isn't the first time I've run into this exact class of issue and I don't expect it to be the last unless something like this is implemented

PR #193, opened 2019-11-27
Use v1.CSINode in the external-attacher

### LINKED ISSUE #197: Update external-attacher to use v1.CSINode
We can also remove the fallback on Node annotation at the same time. This change will require a major version bump because we would have to bump the min K8s version to 1.17 when the object was introduced.

We should make this change at least 1, if not 2 releases before the v1beta gets removed from Kubernetes.

/lifecycle frozen

bertinatto:
**What type of PR is this?**

/kind feature

This PR changes to attacher to use the GA version of the CSINode object.

Fixes #197

```release-note
Use GA version of CSINode object.
```

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -155,7 +155,7 @@ func main() {
 		if supportsAttach {
 			pvLister := factory.Core().V1().PersistentVolumes().Lister()
 			vaLister := factory.Storage().V1().VolumeAttachments().Lister()
-			csiNodeLister := factory.Storage().V1beta1().CSINodes().Lister()
+			csiNodeLister := factory.Storage().V1().CSINodes().Lister()
 			volAttacher := attacher.NewAttacher(csiConn)
 			CSIVolumeLister := attacher.NewVolumeLister(csiConn)
 			handler = controller.NewCSIHandler(clientset, csiAttacher, volAttacher, CSIVolumeLister, pvLister, csiNodeLister, vaLister, timeout, supportsReadOnly, csitrans.New())
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -32,7 +32,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	storagelisters "k8s.io/client-go/listers/storage/v1"
-	storagelistersv1beta1 "k8s.io/client-go/listers/storage/v1beta1"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 )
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -8,66 +8,63 @@ require (
...
-	k8s.io/csi-translation-lib v0.0.0-20191009030015-17db17aaadeb
+	k8s.io/api v0.0.0-20191122220107-b5267f2975e0
+	k8s.io/apimachinery v0.0.0-20191121175448-79c2a76c473a
+	k8s.io/client-go v0.0.0-20191122220542-ed16ecbdf3a0
+	k8s.io/csi-translation-lib v0.0.0-20191122224157-518651485193
 	k8s.io/klog v1.0.0
 )
 
-replace k8s.io/api => k8s.io/api v0.0.0-20190918195907-bd6ac527cfd2
+replace k8s.io/api => k8s.io/api v0.0.0-20191122220107-b5267f2975e0
```
pohly: Which version of Kubernetes is that? Can you record that in the commit message?

I assume all code coming from Kubernetes is at that version?

msau42: I vaguely remember seeing a script that can update all k8s dependencies. Was that something you wrote?

cmd/csi-attacher/main.go
```diff
@@ -147,7 +147,7 @@ func main() {
 			pvLister := factory.Core().V1().PersistentVolumes().Lister()
 			nodeLister := factory.Core().V1().Nodes().Lister()
 			vaLister := factory.Storage().V1beta1().VolumeAttachments().Lister()
```
msau42: Hm we should use v1 VolumeAttachment too.

go.mod
pohly: I took some snippet that was posted in one of the Kubernetes discussions around semantic versioning and added it to csi-release-tools, so it's now in all of our repos: https://github.com/kubernetes-csi/csi-release-tools/blob/master/go-get-kubernetes.sh

But it's impossible to tell from the result whether that script was used and thus whether the new replacements are consistent. We basically have to trust the PR submitter that they did the right thing. To reproduce it, the version given to the scripts needs to be documented, hence me question about the commit message.

bertinatto: > Which version of Kubernetes is that? Can you record that in the commit message?

It was `v1.17.0-rc.1`, but today I updated it to `v1.17.0-rc.2` (which was released yesterday). I also updated the commit message to record both the k8s version and how the dependencies were updated (using the `go-get-kubernetes.sh` script).

cmd/csi-attacher/main.go
bertinatto: I can create another PR to address that...

bertinatto:
/hold

Checking how to support older k8s releases

jsafrane:
@bertinatto, with Kubernetes 1.19, IMO we don't need to worry about supporting v1beta1 of CSINode. We can require Kubernetes 1.17. Can you please continue with this PR?

bertinatto:
> @bertinatto, with Kubernetes 1.19, IMO we don't need to worry about supporting v1beta1 of CSINode. We can require Kubernetes 1.17. Can you please continue with this PR?

Done, rebased PR. I think it's ready for review.

PR #199, opened 2019-12-21
Prepare release 2.1

jsafrane:
/kind documentation

Update README and CHANGELOG for release 2.1

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ The following table reflects the head of this branch.
 | Feature       | Status  | Default | Description                                                                                   |
 | ------------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
 | CSINode*      | Beta    | On      | external-attacher uses the CSINode object to get the driver's node name instead of the Node annotation. |
-| CSIMigration* | Alpha   | On      | [Migrating in-tree volume plugins to CSI](https://kubernetes.io/docs/concepts/storage/volumes/#csi-migration). |
+| CSIMigration* | Beta    | On      | [Migrating in-tree volume plugins to CSI](https://kubernetes.io/docs/concepts/storage/volumes/#csi-migration). |
 
 *) There are no special feature gates for these features. They are enabled by turning on the corresponding features in Kubernetes.
 
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@ The external-attacher is an external controller that monitors `VolumeAttachment`
 
 This information reflects the head of this branch.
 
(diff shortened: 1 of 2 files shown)


jsafrane:
cc @msau42, PTAL

README.md
```diff
@@ -21,7 +21,7 @@ This information reflects the head of this branch.
 
 | Compatible with CSI Version                                                                | Container Image             | Min K8s Version |
 | ------------------------------------------------------------------------------------------ | ----------------------------| --------------- |
-| [CSI Spec v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | quay.io/k8scsi/csi-attacher | 1.15            |
+| [CSI Spec v1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0) | quay.io/k8scsi/csi-attacher | 1.15            |
```
msau42: Should we add recommended k8s version column since migration is beta in 1.17?

README.md
```diff
@@ -32,7 +32,7 @@ The following table reflects the head of this branch.
 | Feature       | Status  | Default | Description                                                                                   |
 | ------------- | ------- | ------- | --------------------------------------------------------------------------------------------- |
 | CSINode*      | Beta    | On      | external-attacher uses the CSINode object to get the driver's node name instead of the Node annotation. |
-| CSIMigration* | Alpha   | On      | [Migrating in-tree volume plugins to CSI](https://kubernetes.io/docs/concepts/storage/volumes/#csi-migration). |
+| CSIMigration* | Beta    | On      | [Migrating in-tree volume plugins to CSI](https://kubernetes.io/docs/concepts/storage/volumes/#csi-migration). |
```
msau42: CSINode is GA now, but we still use beta object for backwards compatibility. How do you want to note it? Or can we remove it altogether? Our min version is 1.15 which means that it should not be falling back on node annotation anymore

jsafrane: Have we ever removed `storage.k8s.io/v1beta1`? IMO we've always added new objects there (+v1alpha1) and we do not deprecate / remove it. We can keep using it and move to v1 in the next release, bumping just min. k8s version.

We should not use the node annotation, I filed https://github.com/kubernetes-csi/external-attacher/issues/203.

README.md
jsafrane: added recommended version

jsafrane:
Added recommended version + Saad's metrics to CHANGELOG

PR #213, opened 2020-01-31
Removed dependency on node annotation in favor of CSINode resource

### LINKED ISSUE #203: Stop using node annotation
Since the external-attacher uses CSINode object that has reached beta in all supported Kubernetes releases, the attacher should stop using `csi.volume.kubernetes.io/nodeid` annotation on Node objects.

Danil-Grigorev:
**What type of PR is this?**

/kind cleanup

CSINode is in all supported Kubernetes releases, so no more Node annotation is required.

Fixes #203 

```release-note
Removed usage of annotation csi.volume.kubernetes.io/nodeid on Node objects. The external-attacher now requires Kubernetes 1.14 with feature gate CSINodeInfo enabled.
```

### DIFF
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -194,17 +194,8 @@ func node() *v1.Node {
 	return &v1.Node{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: testNodeName,
-			Annotations: map[string]string{
-				"csi.volume.kubernetes.io/nodeid": fmt.Sprintf("{ %q: %q }", testAttacherName, testNodeID),
-			},
 		},
 	}
-}
-
-func nodeWithoutAnnotations() *v1.Node {
-	n := node()
-	n.Annotations = nil
-	return n
 }
 
 func csiNode() *storage.CSINode {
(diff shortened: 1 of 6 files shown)


msau42:
Thanks for the PR! Can you work on resolving the CLA issues?

msau42:
Nm looks like CLA issue has been resolved

/ok-to-test

jsafrane:
@Danil-Grigorev, thanks for the PR. Can you please edit the release note in the first comment? This is a change visible to user and they should be warned.

Something like:
`Removed usage of annotation csi.volume.kubernetes.io/nodeid on Node objects. The external-attacher now requires Kubernetes 1.14 with feature gate CSINodeInfo enabled.`

pkg/controller/csi_handler_test.go
```diff
@@ -832,28 +823,28 @@ func TestCSIHandler(t *testing.T) {
 			},
 		},
 		{
-			name:           "Node without annotations -> error",
-			initialObjects: []runtime.Object{pvWithFinalizer(), nodeWithoutAnnotations()},
+			name:           "Node without CSINode -> error",
```
jsafrane: Can you please add test case where Node has the annotation and CSINode does not exists? It should fail, same as this tests. Just to make sure we really do not process the annotation accidentally.

Danil-Grigorev: Sure, will check.

jsafrane:
/lgtm
/approve

Thanks for your PR!

PR #218, opened 2020-03-25
Attacher reacts for deleted PVs if finalizer is present

### LINKED ISSUE #217: Unable to release finalizer for PVs created via legacy storage class
The out-of-tree CSI drivers are still ([and forever?](https://kubernetes.slack.com/archives/CG04EL876/p1578500027013800?thread_ts=1578474999.005000&cid=CG04EL876)) compatible with the legacy (in-tree) provisioners, e.g. `kubernetes.io/aws-ebs`.
Provisioning + attaching a PV referencing such a storage class works perfectly fine with solely CSI. However, deleting such a volume is buggy as the external-attacher does not release its finalizer on the related `PersistentVolume` object.

It only watches for all `VolumeAttachment` and adds its finalizer to the referenced `PersistentVolume` on creation/update. However, when the PV gets deleted it checks whether `.spec.csi` is set which is not the case for a PV using a legacy provisioner. It correctly detaches the volume and releases the `VolumeAttachment`, though, the `PersistentVolume` will stuck forever in the system.

The culprit function seems to be https://github.com/kubernetes-csi/external-attacher/blob/2d2ab15c96d6b22e9fb7c290b22bafa8d2cee4e1/pkg/controller/controller.go#L219-L224

/kind bug

rfranzke:
**What type of PR is this?**
/kind bug

When a `PersistentVolume` gets deleted the external-attacher should consider it in case it has previously added its finalizer to it. Otherwise (for create/update operations), it can continue to only consider it if `.spec.csi != nil`.

Fixes #217

NONE

```release-note
A bug that prevented the external-attacher from releasing its finalizer from a `PersistentVolume` object that was created using a legacy storage class provisioner has been fixed.
```

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -20,12 +20,11 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/klog"
-
 	v1 "k8s.io/api/core/v1"
 	storage "k8s.io/api/storage/v1beta1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	storageinformers "k8s.io/client-go/informers/storage/v1beta1"
@@ -37,6 +36,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
 )
 
 // CSIAttachController is a controller that attaches / detaches CSI volumes using provided Handler interface
@@ -217,10 +217,7 @@ func (ctrl *CSIAttachController) syncVA() {
 }
 
 func (ctrl *CSIAttachController) processFinalizers(pv *v1.PersistentVolume) bool {
-	if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == ctrl.attacherName {
-		return true
-	}
-	return false
+	return pv.DeletionTimestamp != nil && sets.NewString(pv.Finalizers...).Has(GetFinalizerName(ctrl.attacherName))
 }
 
 // syncPV deals with one key off the queue.  It returns false when it's time to quit.
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -1198,6 +1198,26 @@ func TestCSIHandler(t *testing.T) {
 			},
 		},
 		{
+			name:           "VA deleted -> PV finalizer removed (GCE PD PV)",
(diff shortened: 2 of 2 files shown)


pkg/controller/controller.go
```diff
@@ -217,10 +217,10 @@ func (ctrl *CSIAttachController) syncVA() {
 }
 
 func (ctrl *CSIAttachController) processFinalizers(pv *v1.PersistentVolume) bool {
-	if pv.Spec.CSI != nil && pv.Spec.CSI.Driver == ctrl.attacherName {
-		return true
+	if pv.DeletionTimestamp != nil {
+		return sets.NewString(pv.Finalizers...).Has(GetFinalizerName(ctrl.attacherName))
 	}
-	return false
+	return pv.Spec.CSI != nil && pv.Spec.CSI.Driver == ctrl.attacherName
```
jsafrane: I think you can safely remove this `pv.Spec.CSI` check now.

rfranzke: Hm, wouldn't the attacher then also consider `PersistentVolume`s whose `.spec.csi.driver` doesn't match to its own attacher name? I guess this was the reason why the check was there in the first place?

jsafrane: The finalizer name is derived from CSI driver name, so IMO it's not necessary to check `pv.spec.csi.driver`.

rfranzke: OK. I'm not too deep in the details, but got this question: How would a setup with multiple CSI drivers in the same cluster look like? Would I have only one instance of the external-attacher or one per CSI driver?

jsafrane: Each driver must have its own external-attacher instance

rfranzke: OK, thanks. Suppose the discussed check gets removed - wouldn't these multiple external-attacher instances then all start to reconcile the `PersistentVolume` although at least some of them are not responsible?

jsafrane: That should be filtered by finalizer check you added above - each attacher will process only PVs with its own `GetFinalizerName(ctrl.attacherName))`.

Oh, btw, the function should return `false` if deletion timestamp is not set, of course.

jsafrane:
@rfranzke, are you still working on the fix? IMO it's very close to get merged.

rfranzke:
Yes, sorry, I was busy the last days. I'll finish it tonight.

rfranzke:
Can you please check again? I was not entirely sure what you meant with

> can you please add an unit test for a migrated volume? Something like this, but with a migrated PV (so create migratedPVWithFinalizer()):

Do you consider a migrated PV a PV with the annotation? If so, how is it related to the change? Instead, I added three test cases for the finalizer removal, please check if this is what you meant.

jsafrane:
I did not notice there already is `gcePDPVWithFinalizer`. You got it right :-).

/lgtm
/approve

rfranzke:
Thanks for approving! Do you want me to open a cherry-pick PR for the [`release-2.1`](https://github.com/kubernetes-csi/external-attacher/tree/release-2.1) branch?

PR #227, opened 2020-04-22
Update VolumeAttachment.Finalizers as separate subresource in markAsDetatched

### LINKED ISSUE #225: Failed to remove finalizer on VolumeAttachment
After https://github.com/kubernetes-csi/external-attacher/pull/200 merged, our pd driver tests started failing/flaking, especially the test cases that create a 2nd pod on the same node.

It seems like the attacher keeps trying to processing the VolumeAttachment object, but VolumeAttachment object is not able to be deleted. These messages are repeated forever:
```
1124 I0420 21:40:11.832369       1 csi_handler.go:549] Detached "csi-4b1f34f83ecd1528d4aafd1104099e9aaf3e0d81222ab41ccf999555c3c1ff7b"
1125 I0420 21:40:11.832606       1 util.go:80] Marking as detached "csi-4b1f34f83ecd1528d4aafd1104099e9aaf3e0d81222ab41ccf999555c3c1ff7b"
1126 I0420 21:40:11.837749       1 util.go:95] Finalizer removed from "csi-4b1f34f83ecd1528d4aafd1104099e9aaf3e0d81222ab41ccf999555c3c1ff7b"
1127 I0420 21:40:11.837787       1 csi_handler.go:280] Fully detached "csi-4b1f34f83ecd1528d4aafd1104099e9aaf3e0d81222ab41ccf999555c3c1ff7b"
1128 I0420 21:40:11.837817       1 csi_handler.go:225] CSIHandler: finished processing "csi-4b1f34f83ecd1528d4aafd1104099e9aaf3e0d81222ab41ccf999555c3c1ff7b"

```

mattcary:
/kind bug

The update of the VolumeAttachment version from v1beta to v1 changed its status to be a subresource. markAsDetached() updated both the status and the finalizers in a single patch, and the update to v1 blindly put in a "status" subresource restriction, causing the finalizer removal to be silently ignored. This change removes the finalizer in a second patch.

Fixes #225

```release-note
Fixed #225
```

### DIFF
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -919,8 +919,11 @@ func TestCSIHandler(t *testing.T) {
 			addedVA:        deleted(va(true, fin, ann)),
 			expectedActions: []core.Action{
 				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
-					types.MergePatchType, patch(deleted(va(true, fin, ann)),
+					types.MergePatchType, patch(deleted(va(true, "", ann)),
 						deleted(va(false /*attached*/, "", ann))), "status"),
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(deleted(va(false, fin, ann)),
+						deleted(va(false /*attached*/, "", ann)))),
 			},
 			expectedCSICalls: []csiCall{
 				{"detach", testVolumeHandle, testNodeID, noAttrs, noSecrets, readWrite, success, ignored, noMetadata, 0},
@@ -932,8 +935,11 @@ func TestCSIHandler(t *testing.T) {
 			addedVA:        deleted(vaWithInlineSpec(va(true, fin, ann))),
 			expectedActions: []core.Action{
 				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
-					types.MergePatchType, patch(deleted(vaWithInlineSpec(va(true, fin, ann))),
+					types.MergePatchType, patch(deleted(vaWithInlineSpec(va(true, "", ann))),
 						deleted(vaWithInlineSpec(va(false /*attached*/, "", ann)))), "status"),
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(deleted(vaWithInlineSpec(va(false, fin, ann))),
+						deleted(vaWithInlineSpec(va(false /*attached*/, "", ann))))),
 			},
 			expectedCSICalls: []csiCall{
 				{"detach", testVolumeHandle, testNodeID, noAttrs, noSecrets, readWrite, success, ignored, noMetadata, 0},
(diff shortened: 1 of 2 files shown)


msau42:
Test cases under https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/controller/csi_handler_test.go need to be updated with the additional patch action for removing the finalizer.

I think we can also look into this as a separate followup, but I don't see the https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/controller/framework_test.go checking for volumattachment object deletion. A unit test checking for deletion probably could have caught this issue. cc @jsafrane

mattcary:
Tests are fixed.

For correctly checking that the volumes are detached, would it be enough to just check through the vaInformer that there are no such resources (modulo adding a flag or something for test cases that we expect to leave VAs hanging about)?

msau42:
I'm not sure the fake client/informer is going to handle deletion of the object after we remove all the finalizers. Maybe something we can at least check for in some test case, is that the finalizer is cleared from the object?

pkg/controller/csi_handler_test.go
```diff
@@ -1087,16 +1120,28 @@ func TestCSIHandler(t *testing.T) {
...
+					types.MergePatchType, patch(deleted(va(false, "", ann)),
+						deleted(vaWithDetachError(va(false, "", ann), "persistentvolume \"pv1\" not found"))),
 					"status"),
 				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
-					types.MergePatchType, patch(deleted(va(true, fin, ann)),
-						deleted(vaWithDetachError(va(true, fin, ann), "persistentvolume \"pv1\" not found"))),
+					types.MergePatchType, patch(deleted(va(false, "", ann)),
+						deleted(vaWithDetachError(va(false, "", ann), "persistentvolume \"pv1\" not found"))),
+					"status"),
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
```
msau42, line 1134: This is odd. This test case is supposed to fail the call the first two times, and then succeed on the 3rd.

mattcary: Weird indeed. The test passes with >= 1 actions and only fails when the first action is removed. I'm trying to puzzle out what's going on.

mattcary: Ok, I figured out what was going on with the test. Turns out it wasn't actually testing what it said it was (wrong verb in the reactor), and even then the expected actions weren't behaving as they probably were expected to. I've clarified it all in a comment (or tried to).

msau42: Ok, the test case itself is kind of strange, but I can see the benefit for adding an additional call to verify that we do retry.

My one last nit is that there's actually 5 loops but the comment says 4.

mattcary: d'oh :P

mattcary:
> I'm not sure the fake client/informer is going to handle deletion of the object after we remove all the finalizers. Maybe something we can at least check for in some test case, is that the finalizer is cleared from the object?

So the actions/fake clientset don't actually change any of the resource AFAIU. In order to check that the finalizer is cleared, we'd have to manually apply the patch. However, the Patch Action doesn't seem to expose any subresources, so we couldn't accurately test the main thing we want to test. At least that I can tell. sigh. The idea is close; framework_test.go does exactly that for Updates.

I'm open to further ideas if you have them.

msau42:
Yeah I don't have any great ideas. Opened https://github.com/kubernetes-csi/external-attacher/issues/228 to track investigating possible solutions.

One more last thing, can you squash your commits to 1?

mattcary:
Squashed.

msau42:
/lgtm
/approve
Thanks!

PR #229, opened 2020-04-30
Skip prow job when on Kubernetes v1.16.

bertinatto:
**What type of PR is this?**

/kind cleanup

Going forward we are using the GA version of the CSINode object. However, Kubernetes v1.16 only
contains the v1beta1 version of that object, hence we want to skip the prow job in this version.

```release-note
NONE
```

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -2,4 +2,26 @@
 
 . release-tools/prow.sh
 
+# This check assumes that the current configuration uses a driver deployment
+# which has been updated to use v1 APIs that aren't available in Kubernetes < 1.17.
+# TODO: The check can be removed when all Prow jobs for Kubernetes < 1.17 are removed.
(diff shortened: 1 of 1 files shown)


.prow.sh
```diff
@@ -2,4 +2,26 @@
 
 . release-tools/prow.sh
 
+# This check assumes that the current configuration uses a driver deployment
+# which has been updated to use v1 APIs that aren't available in Kubernetes < 1.17.
+# TODO: The check can be removed when all Prow jobs for Kubernetes < 1.17 are removed.
+if ! (version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || "${CSI_PROW_KUBERNETES_VERSION}" == "latest"); then
```
pohly, line 8: This syntax isn't quite right, leading to:
```
./.prow.sh: line 8: 1.16.2: command not found
```

Has to be:
```
if ! (version_gt "${CSI_PROW_KUBERNETES_VERSION}" "1.16.255" || [ "${CSI_PROW_KUBERNETES_VERSION}" == "latest" ] ); then
```

Issue #228, opened 2020-04-24
Unit tests missing coverage for patch operations


msau42:
Fake client doesn't update informer cache, and applying the patch is hard. Sync up with api-machinery for suggestions.

/kind cleanup

PR #244, opened 2020-08-06
Fixes ReconcileVA for non-migratable drivers

### LINKED ISSUE #243: Is reconciling only available for in-Tree drivers?
Hello, After enabling `csi.ControllerServiceCapability_RPC_LIST_VOLUMES_PUBLISHED_NODES` in controllerserver and including the PublishedNodeIDs field in the ListVolumes response. We started to see the following log message from the csi-attacher sidecar. And I found the following function called as part of the reconciling code from the sidecar:

```
// RepairVolumeHandle generates a correct volume handle based on node ID information.
func (CSITranslator) RepairVolumeHandle(driverName, volumeHandle, nodeID string) (string, error) {
	if plugin, ok := inTreePlugins[driverName]; ok {
		return plugin.RepairVolumeHandle(volumeHandle, nodeID)
	}
	return "", fmt.Errorf("could not find In-Tree driver name for CSI plugin %v", driverName)
}
```
is reconciling not supported for non in-tree CSI drivers? thanks in Advance for your help

yuga711:
**What type of PR is this?**
/kind bug

This PR fixes an issue in the volume attachment reconciler. Currently, the reconciler calls a method (unconditionally) that is supposed to be called only when the corresponding volume driver implements the in-tree plugin. 
The PR also fixes a few existing broken tests.

Fixes #243

Tested the fix with csi-mock driver. The driver does not implement in-tree plugin.
Without the fix, the reconciler failed as below:

```
W0806 02:35:26.595059       1 csi_handler.go:159] Failed to repair volume handle  for driver io.kubernetes.storage.mock: could not find In-Tree driver name for CSI plugin io.kubernetes.storage.mock
```

With the fix, no such failure was observed.

```release-note
Fixes an issue that results in a failure in volume attachment reconciler when the corresponding driver does not implement in-tree plugin. 
```

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -153,10 +153,18 @@ func (h *csiHandler) ReconcileVA() error {
 		}
 		attachedStatus := va.Status.Attached
 
-		volumeHandle, err = h.translator.RepairVolumeHandle(source.Driver, volumeHandle, nodeID)
+		// If volume driver has corresponding in-tree plugin, generate a correct volumehandle
+		isMig, err := h.isMigratable(va)
 		if err != nil {
-			klog.Warningf("Failed to repair volume handle %s for driver %s: %v", volumeHandle, source.Driver, err)
+			klog.Warningf("Failed to check if migratable for volume handle %s (driver %s): %v", volumeHandle, source.Driver, err)
 			continue
+		}
+		if isMig {
+			volumeHandle, err = h.translator.RepairVolumeHandle(source.Driver, volumeHandle, nodeID)
+			if err != nil {
+				klog.Warningf("Failed to repair volume handle %s for driver %s: %v", volumeHandle, source.Driver, err)
+				continue
+			}
 		}
 
 		// Check whether the volume is published to this node
@@ -352,6 +360,25 @@ func (h *csiHandler) hasVAFinalizer(va *storage.VolumeAttachment) bool {
 		}
 	}
 	return false
+}
+
+// Checks if the PV (or) the inline-volume corresponding to the VA could have migrated from
+// in-tree to CSI.
+func (h *csiHandler) isMigratable(va *storage.VolumeAttachment) (bool, error) {
+	if va.Spec.Source.PersistentVolumeName != nil {
+		pv, err := h.pvLister.Get(*va.Spec.Source.PersistentVolumeName)
+		if err != nil {
+			return false, err
+		}
+		return h.translator.IsPVMigratable(pv), nil
+	} else if va.Spec.Source.InlineVolumeSpec != nil {
+		if va.Spec.Source.InlineVolumeSpec.CSI == nil {
+			return false, errors.New("inline volume spec contains nil CSI source")
+		}
+		return true, nil
+	} else {
+		return false, nil
+	}
 }
 
 func getCSISource(pvSpec *v1.PersistentVolumeSpec) (*v1.CSIPersistentVolumeSource, error) {
--- a/pkg/controller/csi_handler_test.go
+++ b/pkg/controller/csi_handler_test.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	core "k8s.io/client-go/testing"
+	csitranslator "k8s.io/csi-translation-lib"
 	"k8s.io/klog/v2"
 )
 
@@ -61,7 +62,7 @@ func csiHandlerFactory(client kubernetes.Interface, informerFactory informers.Sh
 		informerFactory.Storage().V1().VolumeAttachments().Lister(),
 		&timeout,
 		true, /* supports PUBLISH_READONLY */
-		fakeInTreeToCSITranslator{},
+		csitranslator.New(),
 	)
 }
 
@@ -76,7 +77,7 @@ func csiHandlerFactoryNoReadOnly(client kubernetes.Interface, informerFactory in
 		informerFactory.Storage().V1().VolumeAttachments().Lister(),
 		&timeout,
 		false, /* does not support PUBLISH_READONLY */
-		fakeInTreeToCSITranslator{},
+		csitranslator.New(),
 	)
 }
 
@@ -833,8 +834,18 @@ func TestCSIHandler(t *testing.T) {
 				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
 					types.MergePatchType, patch(va(false /*attached*/, "" /*finalizer*/, nil /* annotations */),
 						va(false /*attached*/, fin, ann))),
-				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
-					types.MergePatchType, patch(va(false /*attached*/, fin, ann),
+				// Add finalizer again (see: https://github.com/kubernetes-csi/external-attacher/issues/228)
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone,
+					testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(va(false /*attached*/, fin /*finalizer*/, ann /* annotations */),
+						vaWithAttachError(va(false, fin, ann), "context deadline exceeded")),
+					"status"),
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(va(false /*attached*/, "" /*finalizer*/, nil /* annotations */),
+						va(false /*attached*/, fin, ann))),
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone,
+					testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(vaWithAttachError(va(false, fin, ann), "context deadline exceeded"),
 						va(true /*attached*/, fin, ann)), "status"),
 			},
 			expectedCSICalls: []csiCall{
@@ -1059,6 +1070,10 @@ func TestCSIHandler(t *testing.T) {
 			initialObjects: []runtime.Object{pvWithFinalizer(), csiNode()},
 			addedVA:        deleted(va(true, fin, ann)),
 			expectedActions: []core.Action{
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone,
+					testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(deleted(va(true, "", ann)),
+						deleted(vaWithDetachError(va(true, "", ann), "context deadline exceeded"))), "status"),
 				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
 					types.MergePatchType, patch(deleted(va(true, "", ann)),
 						deleted(va(false /*attached*/, "", ann))), "status"),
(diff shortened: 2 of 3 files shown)


pkg/controller/csi_handler.go
```diff
@@ -354,6 +357,22 @@ func (h *csiHandler) hasVAFinalizer(va *storage.VolumeAttachment) bool {
...
+// in-tree to CSI.
+func (h *csiHandler) isMigratable(va *storage.VolumeAttachment) bool {
+	if va.Spec.Source.PersistentVolumeName != nil {
+		pv, err := h.pvLister.Get(*va.Spec.Source.PersistentVolumeName)
+		if err != nil {
+			return false
+		}
+		return h.translator.IsPVMigratable(pv)
+	} else if va.Spec.Source.InlineVolumeSpec != nil {
+		return va.Spec.Source.InlineVolumeSpec.CSI != nil
```
yuga711, line 370: @msau42: I am not sure if this is the right check for inline vols. Please suggest.

jsafrane: It looks like the right check.

vendor/modules.txt
```diff
@@ -3,10 +3,13 @@ github.com/beorn7/perks/quantile
 # github.com/blang/semver v3.5.0+incompatible
 github.com/blang/semver
 # github.com/container-storage-interface/spec v1.2.0
+## explicit
```
jsafrane, line 6: You must use golang 1.13 to generate vendor. And why is there any vendor change at all? There is no new import in the code.

jsafrane:
Can you please check the unit tests? They test with "com.test.foo" driver. Why did they succeed? Should they be fixed? Or can you  please add a new unit test that covers your code change?

pkg/controller/csi_handler.go
```diff
@@ -354,6 +357,22 @@ func (h *csiHandler) hasVAFinalizer(va *storage.VolumeAttachment) bool {
 	return false
 }
 
+// Checks if the PV (or) the inline-volume corresponding to the VA could have migrated from
+// in-tree to CSI.
+func (h *csiHandler) isMigratable(va *storage.VolumeAttachment) bool {
+	if va.Spec.Source.PersistentVolumeName != nil {
+		pv, err := h.pvLister.Get(*va.Spec.Source.PersistentVolumeName)
+		if err != nil {
+			return false
```
msau42, line 366: should we return error instead of false?

msau42:
Can you update the PR title and description to be more specific to the issue, something like "Fixes ReconcileVA for non-migratable drivers"

vendor/modules.txt
yuga711: Yes, fixed it.

pkg/controller/csi_handler.go
yuga711: Yes, makesense.

yuga711:
> > Can you please check the unit tests? They test with "com.test.foo" driver. Why did they succeed? Should they be fixed? Or can you please add a new unit test that covers your code change?
> 
> I was curious about this too. We seem to already have unit tests for [ReconcileVA](https://github.com/kubernetes-csi/external-attacher/blob/298cb965f909f76f2d2dfa2062cb5e142f14b9d5/pkg/controller/csi_handler_test.go). I think it's because the fake for RepairVolumeHandle [always succeeds](https://github.com/kubernetes-csi/external-attacher/blob/298cb965f909f76f2d2dfa2062cb5e142f14b9d5/pkg/controller/framework_test.go#L528). I wonder though, why do we need a fake translator? Can we use the real one?

Switched to real translator. Added new checks. Fixed several broken existing tests.

pkg/controller/csi_handler.go
```diff
@@ -153,11 +153,19 @@ func (h *csiHandler) ReconcileVA() error {
 		}
 		attachedStatus := va.Status.Attached
 
-		volumeHandle, err = h.translator.RepairVolumeHandle(source.Driver, volumeHandle, nodeID)
+		// If volume driver has corresponding in-tree plugin, generate a correct volumehandle
+		isMig, err := h.isMigratable(va)
 		if err != nil {
-			klog.Warningf("Failed to repair volume handle %s for driver %s: %v", volumeHandle, source.Driver, err)
+			klog.Warningf("Failed to check if migratable: %v", err)
```
msau42, line 159: Can you log the volume handle too?

pkg/controller/csi_handler.go
```diff
@@ -354,6 +362,22 @@ func (h *csiHandler) hasVAFinalizer(va *storage.VolumeAttachment) bool {
...
+// in-tree to CSI.
+func (h *csiHandler) isMigratable(va *storage.VolumeAttachment) (bool, error) {
+	if va.Spec.Source.PersistentVolumeName != nil {
+		pv, err := h.pvLister.Get(*va.Spec.Source.PersistentVolumeName)
+		if err != nil {
+			return false, err
+		}
+		return h.translator.IsPVMigratable(pv), nil
+	} else if va.Spec.Source.InlineVolumeSpec != nil {
+		return va.Spec.Source.InlineVolumeSpec.CSI != nil, nil
```
msau42, line 375: I think it should be an error if InlineVolumeSpec is set, but CSI is nil

pkg/controller/framework_test.go
```diff
@@ -297,6 +302,13 @@ func runTests(t *testing.T, handlerFactory handlerFactory, tests []testCase) {
 		if test.additionalCheck != nil {
 			test.additionalCheck(t, test)
 		}
+		// makesure all the csi calls were executed.
+		if csiConnection.index < len(csiConnection.calls) {
```
msau42, line 306: Is it guaranteed that the test will run the reconciler loop exactly the same number of times?

pkg/controller/csi_handler_test.go
```diff
@@ -34,6 +34,7 @@ import (
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	core "k8s.io/client-go/testing"
+	csitranslator "k8s.io/csi-translation-lib"
```
msau42, line 37: Can we add a test case for ReconcileVA using a volume type that does support migration?

msau42:
Did you try running the fixed tests without your fix to ReconcileVA, to see if the test fails as expected?

pkg/controller/csi_handler.go
yuga711: ack

pkg/controller/framework_test.go
yuga711: I think yes. Otherwise, I think we need to fix such tests.

pkg/controller/csi_handler_test.go
yuga711: [These](https://github.com/kubernetes/kubernetes/blob/ded1f58779d0d2570735a6ebe7428d57a100ca97/staging/src/k8s.io/csi-translation-lib/translate.go#L29) are the ones that truly support in the actual translator. Do you suggest unit tests based on this? (or) we could setup a fake translator that migration support. If it's more work, can we do this separately?

yuga711:
> Did you try running the fixed tests without your fix to ReconcileVA, to see if the test fails as expected?

Yes, the tests fail as expected when I ran them without fix. ie the tests complain saying the expected number of csi calls were not run...

pkg/controller/csi_handler.go
yuga711: ack

pkg/controller/csi_handler_test.go
```diff
@@ -696,7 +697,7 @@ func TestCSIHandler(t *testing.T) {
 		{
 			name:           "unknown node -> error",
 			initialObjects: []runtime.Object{pvWithFinalizer()},
-			addedVA:        va(false, fin, ann),
+			addedVA:        va(false, fin, nil),
```
jsafrane, line 700: Is this change necessary? My tests shows that presence of the annotation does not make any difference, because status is patched, not the object with its finalizers. Having `ann` in the `addedVA` is more correct, as the Patch below expects it's present in the original object.

pkg/controller/csi_handler_test.go
```diff
@@ -833,8 +834,17 @@ func TestCSIHandler(t *testing.T) {
...
 					types.MergePatchType, patch(va(false /*attached*/, "" /*finalizer*/, nil /* annotations */),
 						va(false /*attached*/, fin, ann))),
-				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
-					types.MergePatchType, patch(va(false /*attached*/, fin, ann),
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone,
+					testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(va(false /*attached*/, fin /*finalizer*/, ann /* annotations */),
+						vaWithAttachError(va(false, fin, ann), "context deadline exceeded")),
+					"status"),
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
```
jsafrane, line 842: Please add a comment that the finalizer is added again as a workaround / result of https://github.com/kubernetes-csi/external-attacher/issues/228

pkg/controller/csi_handler_test.go
jsafrane: This applies to all `ann` removed from `addedVA` below

pkg/controller/csi_handler_test.go
```diff
@@ -1381,19 +1395,33 @@ func TestCSIHandler(t *testing.T) {
...
 				va(true /*attached*/, "" /*finalizer*/, nID /*annotations*/),
 				pvWithFinalizer(),
+				csiNode(),
 			},
 			listerResponse: map[string][]string{
 				// Intentionally empty
 			},
+			expectedActions: []core.Action{
+				// Finalizer is saved first
+				core.NewPatchAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
```
jsafrane, line 1418: I think that saving of finalizers has been tested thoroughly in `TestCSIHandler` and here we don't need to test it again. So feel free to set it in `initialObjects`.

pkg/controller/csi_handler_test.go
```diff
@@ -1381,19 +1395,33 @@ func TestCSIHandler(t *testing.T) {
...
+					types.MergePatchType, patch(va(false /*attached*/, "" /*finalizer*/, nil /* annotations */),
+						va(false /*attached*/, fin, nil))),
+				core.NewPatchSubresourceAction(vaGroupResourceVersion, metav1.NamespaceNone, testPVName+"-"+testNodeName,
+					types.MergePatchType, patch(va(false /*attached*/, "", nil),
+						va(false, "", nil)), "status"),
```
jsafrane, line 1423: This change is odd:

* The patch is empty.  My tests show that the attacher indeed executes empty PATCH `{}`. Is there a bug in the attacher?

* While the resulting patch is empty, the way it's constructed is wrong - it expects that the volume gets detached (`va(false, ...`) - should it be `true`? The volume was attached successfully.

pkg/controller/csi_handler_test.go
```diff
@@ -1414,11 +1442,12 @@ func TestCSIHandlerReconcileVA(t *testing.T) {
 		{
 			name: "va not attached actual state attached",
 			initialObjects: []runtime.Object{
-				va(false /*attached*/, "" /*finalizer*/, nID /*annotations*/),
+				deleted(va(false /*attached*/, "" /*finalizer*/, nID /*annotations*/)),
+				// va(false /*attached*/, "" /*finalizer*/, nID /*annotations*/),
```
jsafrane, line 1446: good catch, please remove the commented line

pkg/controller/csi_handler_test.go
msau42: I think adding a unit test using one of the supported translated types would be good.

pkg/controller/csi_handler.go
yuga711: Done

pkg/controller/csi_handler.go
yuga711: Done

pkg/controller/csi_handler_test.go
yuga711: Yes, agreed. Added ann back.

pkg/controller/csi_handler_test.go
yuga711: done

pkg/controller/csi_handler_test.go
yuga711: done

pkg/controller/csi_handler_test.go
yuga711: Haven't found if there is a bug executing empty patch.
Fixed the patch construction logic in the test.

pkg/controller/csi_handler_test.go
yuga711: thanks. done.

pkg/controller/framework_test.go
```diff
@@ -224,6 +224,11 @@ func runTests(t *testing.T, handlerFactory handlerFactory, tests []testCase) {
 					t.Errorf("Failed to reconcile Volume Attachment objects: %v", err)
 				}
 			}
+			if !firstIter && (ctrl.vaQueue.Len() > 0 || ctrl.pvQueue.Len() > 0) {
```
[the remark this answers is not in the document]
yuga711: You are right. Added the iteration logic while experimenting with the tests.

pkg/controller/csi_handler_test.go
```diff
@@ -1381,31 +1396,56 @@ func TestCSIHandler(t *testing.T) {
...
+					types.MergePatchType, patch(va(true /*attached*/, "", nil),
+						va(true, "", nil)), "status"),
+			},
 			expectedCSICalls: []csiCall{
 				{"attach", testVolumeHandle, testNodeID, nil, nil, false, nil, false, nil, 0},
 			},
 		},
+		/* TBD
+		{
+			name:           "GCE PD va attached actual state not attached",
```
jsafrane, line 1428: In general, the patch looks good, is this the only thing missing?

yuga711: Here, I wanted to add a test with one of the supported translator types. Now planning to do that separately. I have created a github issue and added a TODO here.

yuga711: > In general, the patch looks good, is this the only thing missing?
I wanted to add a test with one of the supported translator types. Now planning to do that separately. I have created a github issue and added a TODO here.

PR #277, opened 2020-11-19
Bump csi-lib-utils to v0.9.0

jsafrane:
/kind cleanup

Bump csi-lib-utils to v0.9.0 to get rid of klogv1 and get health check on leader election.

```release-note
* Added leader election health check at the metrics port + path "/healthz/leader-election".
* klog/v2 is used for logging
* process_start_time metric is now reported
```


pohly:
Looks good to me, but I feel that I may have missed some discussion around how to handle metrics and health checks, so I'd rather let someone else approve this.

msau42 (commented):
/assign @verult 

who wanted to also add a new flag and deprecate the metrics-address flag since it's now serving more than metrics. But that could be done as a separate pr

msau42:
Do we want the release note to also mention the other two fixes we're pulling from csi-lib-utils? ie process_start_time metric and klog v2

README.md
```diff
@@ -99,6 +99,13 @@ Correct timeout value depends on the storage backend and how quickly it is able
 
 When CSI driver supports `LIST_VOLUMES` and `LIST_VOLUMES_PUBLISHED_NODES` capabilities, the external attacher periodically syncs volume attachments requested by Kubernetes with the actual state reported by CSI driver. Volumes detached by any 3rd party, but still required to be attached by Kubernetes, will be re-attached back. Frequency of this re-sync is controlled by `--reconcile-sync` command line parameter.
 
+### HTTP endpoint
+
+The external-attacher optionally exposes an HTTP endpont at given address:port specified by `--metrics-address` argument. When enabeled, these two paths are exposed:
+
+* Metrics path, as set by `--metrics-path` argument (default is `/metrics`).
+* Leader election health check at `/healthz/leader-election`. It is recommended to run a liveness probe against this endpoint when leader election is used to kill external-attachers that fail to connect to API server. See https://github.com/kubernetes-csi/csi-lib-utils/issues/66 for details.
```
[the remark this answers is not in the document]
jsafrane: fixed

README.md
```diff
@@ -99,6 +99,13 @@ Correct timeout value depends on the storage backend and how quickly it is able
 
 When CSI driver supports `LIST_VOLUMES` and `LIST_VOLUMES_PUBLISHED_NODES` capabilities, the external attacher periodically syncs volume attachments requested by Kubernetes with the actual state reported by CSI driver. Volumes detached by any 3rd party, but still required to be attached by Kubernetes, will be re-attached back. Frequency of this re-sync is controlled by `--reconcile-sync` command line parameter.
 
+### HTTP endpoint
+
+The external-attacher optionally exposes an HTTP endpont at given address:port specified by `--metrics-address` argument. When enabeled, these two paths are exposed:
```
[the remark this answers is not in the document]
jsafrane: fixed

jsafrane:
Updated the release note too.

cmd/csi-attacher/main.go
```diff
@@ -134,8 +135,17 @@ func main() {
...
 	}
 	klog.V(2).Infof("CSI driver name: %q", csiAttacher)
-	metricsManager.SetDriverName(csiAttacher)
-	metricsManager.StartMetricsEndpoint(*metricsAddress, *metricsPath)
+
+	// Prepare http endpoint for metrics + leader election healthz
+	mux := http.NewServeMux()
+	cmm := metrics.NewCSIMetricsManagerForSidecar(csiAttacher)
+	cmm.RegisterToServer(mux, *metricsPath)
+	go func() {
```
verult, line 143: I'm not sure what the behavior would be to start the server before registering the leader election handler. Might be safer to do it after?

README.md
```diff
@@ -99,6 +99,13 @@ Correct timeout value depends on the storage backend and how quickly it is able
 
 When CSI driver supports `LIST_VOLUMES` and `LIST_VOLUMES_PUBLISHED_NODES` capabilities, the external attacher periodically syncs volume attachments requested by Kubernetes with the actual state reported by CSI driver. Volumes detached by any 3rd party, but still required to be attached by Kubernetes, will be re-attached back. Frequency of this re-sync is controlled by `--reconcile-sync` command line parameter.
 
+### HTTP endpoint
+
+The external-attacher optionally exposes an HTTP endpont at given address:port specified by `--metrics-address` argument. When enabled, these two paths are exposed:
```
verult, line 104: typo: endpont

jsafrane: Fixed

cmd/csi-attacher/main.go
jsafrane: I checked that both paths (metrics + leader check) work. I don't want to move the code further, because there would be two places to start the http server, in `then` and `else` branches of this `if`:

https://github.com/kubernetes-csi/external-attacher/blob/6b94e28d9479e4ebed65b3a373149ec054c5cac7/cmd/csi-attacher/main.go#L196-L205

cmd/csi-attacher/main.go
```diff
@@ -134,8 +135,17 @@ func main() {
 		os.Exit(1)
 	}
 	klog.V(2).Infof("CSI driver name: %q", csiAttacher)
-	metricsManager.SetDriverName(csiAttacher)
-	metricsManager.StartMetricsEndpoint(*metricsAddress, *metricsPath)
+
+	// Prepare http endpoint for metrics + leader election healthz
+	mux := http.NewServeMux()
+	cmm := metrics.NewCSIMetricsManagerForSidecar(csiAttacher)
```
pohly, line 141: While copying this code into external-health-monitor I noticed that `metricsPath` is still allowed to be empty, which means that an if check for that is missing here.

Also, metrics gathering elsewhere instantiates the metrics manager *before* starting to talk to the sidecar and then changes the driver name later - not sure how important that is.

pohly: Actually... I think the metrics data is now always going to be empty because the HTTP server presents data from the `cmm` instance and not the `metricsManager` instance that is used to monitor CSI calls.

pohly: Here's my take on the same update in external-health-monitor: https://github.com/kubernetes-csi/external-health-monitor/pull/46

PR #294, opened 2021-03-12
Process PV finalizer for migrated PV when CSIMigration turned off

Jiawei0227:
**What type of PR is this?**
/kind bug

This PR removes PV external-attacher finalizer when CSIMigration is turned off. Without this change, the in-tree PV controller is not able to delete the PV object when there is no CSI driver in the cluster.

Partially Fixes # https://github.com/kubernetes-csi/external-provisioner/issues/576

This does not fix the issue if the corresponding CSI driver is uninstalled when CSI Migration is turned off. This fix is based on the assumption that CSI driver will still be around for at least a while to remove the finalizer. 

```release-note
Fix a bug where external-attacher finalizer not get lifted for migrated PV when CSIMigration is turned off.
```

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -625,10 +625,29 @@ func (h *csiHandler) SyncNewOrUpdatedPersistentVolume(pv *v1.PersistentVolume) {
 	klog.V(4).Infof("CSIHandler: processing PV %q", pv.Name)
 	// Sync and remove finalizer on given PV
 	if pv.DeletionTimestamp == nil {
-		// Don't process anything that has no deletion timestamp.
-		klog.V(4).Infof("CSIHandler: processing PV %q: no deletion timestamp, ignoring", pv.Name)
-		h.pvQueue.Forget(pv.Name)
-		return
+		ignore := true
+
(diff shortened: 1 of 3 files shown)


pkg/controller/csi_handler.go
```diff
@@ -625,10 +625,25 @@ func (h *csiHandler) SyncNewOrUpdatedPersistentVolume(pv *v1.PersistentVolume) {
...
-		return
+		ignore := true
+
+		// if the PV is migrated this means CSIMigration is disabled so we need to remove the finalizer
+		// to give back the control of the PV to Kube-Controller-Manager
+		if h.translator.IsPVMigratable(pv) {
+			ignore = false
+			if ann := pv.Annotations; ann != nil {
+				if migratedToDriver := ann[annMigratedTo]; migratedToDriver == h.attacherName {
+					ignore = true
```
jsafrane, line 636: It would be perhaps better to log a specific message + call `pvQueue.Forget` and return here. The message logged below is about deletion timestamp.

Jiawei0227: Done. I think the deletion timestamp is related. I added a log to identify the scenario that we need to remove the anno

PR #298, opened 2021-04-07
Rebase to 1.21.0

jsafrane:
/kind cleanup
Update to the latest release-tools and rebase all Kubernetes libs to v1.21.0

```release-note
NONE
```


go.mod
```diff
@@ -18,77 +17,74 @@ require (
...
+	k8s.io/csi-translation-lib v0.21.0
 	k8s.io/klog/v2 v2.8.0
 	k8s.io/kube-openapi v0.0.0-20210305164622-f622666832c1 // indirect
 	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.1.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.1 // indirect
 )
 
-replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.20.0
+replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.21.0
```
msau42, line 38: same question as resize PR. I think we don't need these replace statements if we're not importing k8s.io/kubernetes

pohly: I have less faith in that. But we can try without them and then we'll see whether the default version selection for indirect dependencies will do a good enough job of selecting a consistent set of module versions.

msau42: I tried it out on my local workspace and it seems fine?

jsafrane: Fixed with the new release-tools.

go.mod
```diff
@@ -18,77 +17,32 @@ require (
...
+	k8s.io/csi-translation-lib v0.21.0
 	k8s.io/klog/v2 v2.8.0
 	k8s.io/kube-openapi v0.0.0-20210305164622-f622666832c1 // indirect
 	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.1.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.1.1 // indirect
 )
 
-replace k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.20.0
+replace k8s.io/component-base => k8s.io/component-base v0.21.0
```
msau42, line 38: Even then I think all of these replace statements can be removed.

pohly: Then `go get -u ./...` no longer works. For example, it updates to k8s.io/client-go v1.5.2.

jsafrane: +1 to Patrick, `go mod` is very confused when you remove those replaces.

msau42: thanks. one day I will understand how vendoring works :-)

PR #299, opened 2021-04-09
On missing driver, fall back to volattachment ID

jackkleeman:
**What type of PR is this?**
/kind bug

When the driver is missing in the csinode, that doesn't mean we can't unmap. We should support the volume attachment node ID for unmap. For example CSI might not be running (the node could even be shut down!), so the driver would not be present, but the csi node might exist. I want the behaviour to match the csi node not existing at all; fall back to the node ID in the VA.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/csi_handler.go
+++ b/pkg/controller/csi_handler.go
@@ -742,14 +742,15 @@ func (h *csiHandler) getNodeID(driver string, nodeName string, va *storage.Volum
 			klog.V(4).Infof("Found NodeID %s in CSINode %s", nodeID, nodeName)
 			return nodeID, nil
 		}
-		// CSINode exists, but does not have the requested driver.
-		errMessage := fmt.Sprintf("CSINode %s does not contain driver %s", nodeName, driver)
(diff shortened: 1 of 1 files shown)


pkg/controller/csi_handler.go
```diff
@@ -743,13 +743,11 @@ func (h *csiHandler) getNodeID(driver string, nodeName string, va *storage.Volum
 			return nodeID, nil
 		}
 		// CSINode exists, but does not have the requested driver.
-		errMessage := fmt.Sprintf("CSINode %s does not contain driver %s", nodeName, driver)
-		klog.V(4).Info(errMessage)
-		return "", errors.New(errMessage)
+		err = errors.New(fmt.Sprintf("CSINode %s does not contain driver %s", nodeName, driver))
```
xing-yang, line 746: Can you add more comments here to explain how this could happen like what you said in the PR description?

jackkleeman: done

PR #304, opened 2021-05-06
Panic fix: Handle DeletedFinalStateUnknown in VA deletion

Jiawei0227:
**What type of PR is this?**
/kind bug

external-attacher can panic when there is a DeletedFinalStateUnknown delete event happens. This PR checks the type of obj first before type assert to VA.

Fixes #

```release-note
Fix a bug that the controller can panic crash when it receives DeletedFinalStateUnknown deletion event.
```

/cc @msau42 @jsafrane


jsafrane:
/lgtm
/approve
I think 1-21-on-kubernetes-1-21 is under development right now and that's why it's not required.

PR #308, opened 2021-05-16
Map PV access modes to CSI access modes based on driver capability

chrishenzie:
**What type of PR is this?**
/kind feature

See [KEP#2485](https://github.com/kubernetes/enhancements/pull/2489) for more details.

This PR introduces mappings between Kubernetes PV access modes and CSI access modes. If the CSI driver supports the `SINGLE_NODE_MULTI_WRITER` controller capability, map ReadWriteOnce to `SINGLE_NODE_MULTI_WRITER` and ReadWriteOncePod to `SINGLE_NODE_SINGLE_WRITER`. If the CSI driver does not support this controller capability, map ReadWriteOnce to `SINGLE_NODE_WRITER` (current behavior preserved for backwards compatibility).

Fixes #

The CSI spec change [PR#476 is here](https://github.com/container-storage-interface/spec/pull/476) and just requires merging. Until then, I have a placeholder commit using my forked version of the CSI spec.

Similarly for K8s, the API change and feature addition [PR#102028 is here](https://github.com/kubernetes/kubernetes/pull/102028) and still is in the early stages of review. Until then, I have a placeholder commit using my forked version of the K8s API.

```release-note
Adds mappings for PV access modes to new CSI access modes: `SINGLE_NODE_SINGLE_WRITER` and `SINGLE_NODE_MULTI_WRITER`.
```

/assign @msau42 
/assign @jingxu97


pkg/controller/util.go
```diff
@@ -175,28 +171,79 @@ func GetVolumeCapabilities(pvSpec *v1.PersistentVolumeSpec) (*csi.VolumeCapabili
...
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY, nil
 
 	case m[v1.ReadWriteOnce]:
 		// There is only ReadWriteOnce set
-		cap.AccessMode.Mode = csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER, nil
 
 	default:
-		return nil, fmt.Errorf("unsupported AccessMode combination: %+v", pvSpec.AccessModes)
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("unsupported AccessMode combination: %+v", pvSpec.AccessModes)
```
jsafrane, line 210: Is it OK to return error when the PV supports only `ReadWriteOncePod`? Should it be handled the same as `ReadWriteOnce`?

pkg/controller/util.go
```diff
@@ -175,28 +171,79 @@ func GetVolumeCapabilities(pvSpec *v1.PersistentVolumeSpec) (*csi.VolumeCapabili
...
+	case m[v1.ReadOnlyMany] && m[v1.ReadWriteOncePod]:
+		// This is not possible in the CSI spec
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("CSI does not support ReadOnlyMany and ReadWriteOncePod on the same PersistentVolume")
+
+	case m[v1.ReadOnlyMany]:
+		// There is only ReadOnlyMany set
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY, nil
+
+	case m[v1.ReadWriteOnce]:
+		// There is only ReadWriteOnce set
```
jsafrane, line 237: At this point,can there be both `ReadWriteOnce` and `ReadWriteOncePod` in the PV? Not sure how useful this combination is.

pkg/controller/util.go
chrishenzie: I don't think users should use ReadWriteOncePod if the driver does not support the capability.

pkg/controller/util.go
chrishenzie: Yes this is possible, but ReadWriteOncePod takes precedence since it's more restrictive. Fixed this by enforcing only one allowed to be set.

pkg/controller/util.go
jsafrane: Since most of `ReadWriteOncePod` enhancement is done in the scheduler and kubelet to make sure a volume is not presented to multiple pods on the same node, IMO it makes sense to fall back to `ReadWriteOnce` behavior during attach and maybe in kubelet too. The new access mode will be hard to use, because only some CSI drivers will support it, at least initially. Application writers cannot write e.g. a helm chart that reliably deploys their apps on a generic cluster with `ReadWriteOncePod`, because they don't know what CSI drivers are deployed on a random cluster.

cmd/csi-attacher/main.go
```diff
@@ -198,13 +198,31 @@ func main() {
 			klog.Error(err.Error())
 			os.Exit(1)
 		}
+
+		supportsSingleNodeMultiWriter, err := supportsSingleNodeMultiWriter(ctx, csiConn)
```
jsafrane, line 202: can you please add it into `supportsControllerPublish()` above (and maybe rename it), so the attacher calls `GetControllerCapabilities` only once?

vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go
```diff
@@ -0,0 +1,110 @@
...
+
+	default:
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("unsupported AccessMode combination: %+v", pvAccessModes)
+	}
+}
+
+// ToSingleNodeMultiWriterCapableCSIAccessMode maps PersistentVolume access
+// modes in Kubernetes to CSI access modes for drivers that support the
+// SINGLE_NODE_MULTI_WRITER capability.
+func ToSingleNodeMultiWriterCapableCSIAccessMode(pvAccessModes []v1.PersistentVolumeAccessMode) (csi.VolumeCapability_AccessMode_Mode, error) {
```
jsafrane, line 69: Just wondering, would it be better to have one function with an extra argument `ToCSIAccessMode(..., singleNodeMultiWriterSupported bool)`? With two functions, all callers must do the same
```
if singleNodeMultiWriterSupported { 
   am, err = ToSingleNodeMultiWriterCapableCSIAccessMode(...)
} else {
   am, err= ToCSIAccessMode(...)
}
```

jsafrane (commented):
I have couple of nits, but the PR goes the right direction.

vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go
chrishenzie: I think having a boolean argument for this is a bit of a code smell but since we're only dealing with one it's ok. I'll make this change in the library. Also assigned you to the PR for that, https://github.com/kubernetes-csi/csi-lib-utils/pull/89

cmd/csi-attacher/main.go
chrishenzie: I consolidated this logic into one call and put it in a separate commit, PTAL

chrishenzie:
/hold

Rebased on #317 which is blocked on release of v1.22.

chrishenzie:
@jsafrane PTAL, this is ready for review

PR #311, opened 2021-06-10
make use of ratelimiter in framwork test

humblec:
At present the framework test does not make use of the ratelimiter
with `retryInterval{start/max} parameters. This replaces the
framework test with the default values of these params

```release-note-none

```


pohly:
You description says what you are doing, but not why. Can you add that to the commit and PR description?

jsafrane:
+1 to @pohly , what problem this PR solves? What's wrong with `DefaultControllerRateLimiter`?

humblec:
@pohly @jsafrane my intention was to make use of the default values in the test too as these values are different compared to default rate limiter. But it looks like its not needed and its good to keep the default ratelimiter in the test. In that context let me close this PR. Thanks for the thoughts!

PR #331, opened 2021-11-04
Add TLS support for diagnostics HTTP server

VirrageS:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: It adds TLS support for diagnostics HTTP server. In our cluster we require the metrics to be exposed as HTTPS. Not having this support stops us from using metrics from `external-attacher`.

I'm not fully sure about the naming of the flags. I've looked at https://kubernetes.io/docs/setup/best-practices/certificates/ but there isn't seem to be consensus about the flag naming for TLS files. 

**Special notes for your reviewer**: NONE

```release-note
Add TLS support for diagnostics/metrics/leader election health status HTTP server by adding new flags: `--http-endpoint-key-file` and `--http-endpoint-cert-file`.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,9 +67,13 @@ Note that the external-attacher does not scale with more replicas. Only one exte
 
 * `--retry-interval-max`: The exponential backoff maximum value. See [CSI error and timeout handling](#csi-error-and-timeout-handling) for details. 5 minutes is used by default.
 
-* `--http-endpoint`: The TCP network address where the HTTP server for diagnostics, including metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
+* `--http-endpoint`: The TCP network address where the HTTP(S) server for diagnostics, including metrics and leader election health check, will listen (example: `:8080` which corresponds to port 8080 on local host). The default is empty string, which means the server is disabled.
 
 * `--metrics-path`: The HTTP path where prometheus metrics will be exposed. Default is `/metrics`.
+
+* `--http-endpoint-key-file`: The path to TLS key file used for running HTTPS server. The default is empty string, which means that TLS will not be used when running HTTP server. Both `--http-endpoint-key-file` and `--http-endpoint-cert-file` flags must be set to enable TLS support.
+
+* `--http-endpoint-cert-file`: The path to TLS cert file used for running HTTPS server. The default is empty string, which means that TLS will not be used when running HTTP server. Both `--http-endpoint-key-file` and `--http-endpoint-cert-file` flags must be set to enable TLS support.
 
 * `--reconcile-sync`: Resync frequency of the attached volumes with the driver. See [Periodic re-sync](#periodic-re-sync) for details. 1 minute is used by default.
 
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -68,9 +68,11 @@ var (
 
 	reconcileSync = flag.Duration("reconcile-sync", 1*time.Minute, "Resync interval of the VolumeAttachment reconciler.")
 
-	metricsAddress = flag.String("metrics-address", "", "(deprecated) The TCP network address where the prometheus metrics endpoint will listen (example: `:8080`). The default is empty string, which means metrics endpoint is disabled. Only one of `--metrics-address` and `--http-endpoint` can be set.")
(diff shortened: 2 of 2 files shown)


VirrageS:
I signed it

VirrageS:
I signed it

VirrageS:
I signed it

VirrageS:
I signed it

VirrageS:
I signed it

jsafrane:
What people in my social bubble use is [kube-rbac-proxy](https://brancz.com/2018/02/27/using-kube-rbac-proxy-to-secure-kubernetes-workloads/). I.e. the attacher (and other sidecars) listen on http on loopback only and another sidecar listen on a public https port and does https -> http proxying + authentication + authorization (not everyone can read metrics, right?)

It leads to yet another explosion of sidecars - now each sidecar needs its own kube-rbac-proxy container, maybe there is some room for improvement. But I personally don't like adding random https support to CSI sidecars. With this PR we get authentication (both metrics client and the attacher can trust each other who they are), but it does not add any authorization nor audit.

VirrageS:
The problem is that we decided to move away from `kube-rbac-proxy`. I don't know the details but it's just not feasible solution for us.

Therefore, the way to go is to allow exposing the HTTPS for the metrics.

> But I personally don't like adding random https support to CSI sidecars.

Could you explain how this is random?

> With this PR we get confidentiality and authentication (both metrics client and the attacher can trust each other who they are), but it does not add any authorization nor audit.

Yes but I don't see what's the problem with that. It solves one problem. The authorization can be done independently.

Jiawei0227:
> both metrics client and the attacher can trust each other who they are

I think this is only server side verification, i.e., only client trust this is the right server. But we did not add client-side verification in this PR which let the server to trust the client that tries to connect to the HTTPs server

jsafrane:
> The problem is that we decided to move away from kube-rbac-proxy. I don't know the details but it's just not feasible solution for us.

Can you please check what is wrong with kube-rbac-proxy? Maybe it could be fixed there.

>> But I personally don't like adding random https support to CSI sidecars.

> Could you explain how this is random?

It implements only part of TLS, (server) authentication and confidentiality. It leaves authorization and audit open. Why do you want to encrypt the metrics if anyone can get them anyway because of no client authn+authz?

> Yes but I don't see what's the problem with that. It solves one problem. The authorization can be done independently.

How can be authorization added later? With my downstream hat on, it's part of security, not afterthought. Maybe I am too paranoid.

pohly:
> > Yes but I don't see what's the problem with that. It solves one problem. The authorization can be done independently.
> 
> How can be authorization added later? With my downstream hat on, it's part of security, not afterthought. Maybe I am too paranoid.

You aren't. I myself am more worried about unauthorized access to metrics data (not addressed by this PR) than I am about a man-in-the-middle attack where someone fakes metrics data (addressed by this PR via authentication of the HTTPS server and encryption).

VirrageS:
> Can you please check what is wrong with kube-rbac-proxy? Maybe it could be fixed there.

I don't think there is anything wrong with the `kube-rbac-proxy`. It's just we don't want to use another sidecar if we can do it natively (it saves memory and CPU).

> It implements only part of TLS, (server) authentication and confidentiality. It leaves authorization and audit open. Why do you want to encrypt the metrics if anyone can get them anyway because of no client authn+authz?

Because in our use-case we can ensure that the client is authorized but we still want authentication.

> How can be authorization added later? With my downstream hat on, it's part of security, not afterthought. Maybe I am too paranoid.

Hm, maybe I'm missing something but why cannot we? For me, adding authz is just work for another MR.

Authn and authz feel like the two separate use-cases. I agree that ideally we could have support for both of them. But at the same time I don't see a reason why cannot it be added later. If someone would like to have this feature (authz) then I think there will be another MR.

> You aren't. I myself am more worried about unauthorized access to metrics data (not addressed by this PR) than I am about a man-in-the-middle attack where someone fakes metrics data (addressed by this PR via authentication of the HTTPS server and encryption).

Feel free to do MR follow up :))

VirrageS:
Any update on if it's going to be merged?

VirrageS:
Bump

VirrageS:
@jsafrane @xing-yang  is there anything that can be done here to make this mergable?

jsafrane:
> is there anything that can be done here to make this mergable?

We're not going to merge half baked TLS without proper authorization support. It protects metrics from MITM attacks, but the man in the middle can scrape the metrics by themeselves by themselves, so the encryption does not help in any way.

PR #342, opened 2022-01-05
allow setting default fs type through command line argument

### LINKED ISSUE #287: external-attacher should not decide default filesystem type
When filesystem type (FSType) is empty, the csi-attacher sidecar changes it to "ext4" before calling the CSI driver's ControllerPublishVolume().

It should not. While the default for many types of volume is "ext4", no default is specified for a CSI volume, and that is correct: it is the particular CSI driver that must decide what the default filesystem is when none is specified, not the sidecar container.

There appears to be no simple workaround: a driver using external-attacher has no way of knowing whether the "ext4" it sees in a ControllerPublishVolume is there because it was specifically requested or because the csi-attacher has exceeded its authority.

Therefore remove `defaultFSType` and the associated code from external-attacher.

After this fix, some existing CSI drivers may need to change to preserve their existing response to a PVC with an unspecified filesystem. If that is a significant concern then consider replacing defaultFSType with a command-line argument with a default value of "ext4" but which can at least be set to "" to disable the current interference. However, simple removal of defaultFSType is the preferable fix.

/kind bug

RomanBednar:
The external attacher was using a hardcoded default ext4 filesystem which was applied when PV spec did not contain fstype.

As a result attacher always reported ext4 as fs_type volume capability creating interference with some CSI drivers.

This patch adds a flag to csi-attacher binary which provides users with a workaround to set their own default filesystem or simply disable the interference by using an empty string:

`
./bin/csi-attacher --default-fstype ""
`

When the flag is not used  we still use ext4, next major release will allow behavior change and we can default to "" instead preserving the argument in case it's ever needed.

/kind bug

Fixes #287 

```release-note
There is a new --default-fstype flag available in this release that defaults to ext4 and can be used to set any value a driver may need. In the next major release ext4 as the default fsType will be deprecated and replaced by ""
```

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -108,7 +108,6 @@ func markAsDetached(client kubernetes.Interface, va *storage.VolumeAttachment) (
 }
 
 const (
-	defaultFSType      = "ext4"
 	vaNodeIDAnnotation = "csi.alpha.kubernetes.io/node-id"
 )
 
@@ -140,7 +139,7 @@ func GetNodeIDFromCSINode(driver string, csiNode *storage.CSINode) (string, bool
 
 // GetVolumeCapabilities returns a VolumeCapability from the PV spec. Which access mode will be set depends if the driver supports the
 // SINGLE_NODE_MULTI_WRITER capability.
-func GetVolumeCapabilities(pvSpec *v1.PersistentVolumeSpec, singleNodeMultiWriterCapable bool) (*csi.VolumeCapability, error) {
+func GetVolumeCapabilities(pvSpec *v1.PersistentVolumeSpec, singleNodeMultiWriterCapable bool, defaultFSType string) (*csi.VolumeCapability, error) {
 	if pvSpec.CSI == nil {
 		return nil, errors.New("CSI volume source was nil")
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/util.go
```diff
@@ -158,6 +157,8 @@ func GetVolumeCapabilities(pvSpec *v1.PersistentVolumeSpec, singleNodeMultiWrite
 		fsType := pvSpec.CSI.FSType
 		if len(fsType) == 0 {
 			fsType = defaultFSType
+			klog.Warningf("Filesystem type not found in PV spec. Will use fstype set by attacher: %v. " +
+				"This will be deprecated in future versions and defaults will be set by CSI driver.", fsType)
```
jsafrane, line 161: I think we will not deprecate `--default-fstype` parameter, we will change its default to "". IMO the message should be at lower log level (it can be logged quite a lot) and just note that it's using `defaultFSType` as the fsType.

jsafrane (commented):
Please update also README.md and add a release note to the inital comment of this PR - this is an user facing change.
It should clearly say that ext4 as the default fsType is deprecated and will be replaced by `""` in the next major release. And there is `--default-fstype` available in this release that can be used to set it explicitly to any value a driver may need.

RomanBednar:
i signed it

RomanBednar:
I signed it

PR #354, opened 2022-05-16
adjust the RBAC to only have PATCH verb

humblec:
Considering Update() are replaced by Patch() call, we no longer
need the UPDATE RBAC for the sidecar.

> /kind cleanup

```release-note
NONE
```

### DIFF
--- a/deploy/kubernetes/rbac.yaml
+++ b/deploy/kubernetes/rbac.yaml
@@ -69,7 +69,7 @@ metadata:
 rules:
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
-  verbs: ["get", "watch", "list", "delete", "update", "create"]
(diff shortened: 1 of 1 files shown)


jsafrane:
If I am looking right, lease based leader election still does `Update()`, https://github.com/kubernetes-csi/external-attacher/blob/5545207e42af726eeb6b351bab24d7f6e11d3efb/vendor/k8s.io/client-go/tools/leaderelection/resourcelock/leaselock.go#L75

It seems to me that there is no leader election configured in our CI, as it should fail with this PR.

humblec:
>If I am looking right, lease based leader election still does Update(), 
> It seems to me that there is no leader election configured in our CI, as it should fail with this PR.

True.. considering that, this has to be on hold or has to be closed for now..  I am more inclined to close this PR though.

pohly:
I also prefer to close it.

PR #389, opened 2022-11-22
Update K8s dependency go modules for K8s v1.26.0-rc.0

sunnylovestiramisu:
Ran 
```
release-tools/go-get-kubernetes.sh -p 1.26.0-rc.0
```


msau42:
Did the script generate the PR title? I would probably clarify that this is updating the Kubernetes dependencies

sunnylovestiramisu:
> 

Yes it does, we can specify whatever title we want.

humblec:
@sunnylovestiramisu Thanks for the PR.. but few comments from my end. imo, we dont bring `rc` updates to the repo, atleast released minor versions. This script also require some cleaning after getting the kube versions..etc, for ex: the unwanted ` replace` directives. Also considering we have automated this process of updating the dependencies via `dependabot`, this PR shouldnt be necessary.

sunnylovestiramisu:
> @sunnylovestiramisu Thanks for the PR.. but few comments from my end. imo, we dont bring `rc` updates to the repo, atleast released minor versions. This script also require some cleaning after getting the kube versions..etc, for ex: the unwanted ` replace` directives. Also considering we have automated this process of updating the dependencies via `dependabot`, this PR shouldnt be necessary.

So the dependabot has already covered both the script release-tools/update-vendor.sh and release-tools/go-get-kubernetes.sh? Then we should update the documentation: [csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md](https://github.com/kubernetes-csi/csi-release-tools/blob/master/SIDECAR_RELEASE_PROCESS.md) ?

msau42:
@humblec I'm not sure if dependabot can handle kubernetes version updating (at least I have seen it https://github.com/kubernetes-csi/external-provisioner/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+is%3Aclosed).

msau42:
In terms of whether we should wait for 1.26.0, I think it's fine to update to rc.0 now, and then update once more to 1.26.0 before we do the final release. That way we can start running tests with the new deps sooner. I don't expect signficant changes between rc.0 and .0

humblec:
> @humblec I'm not sure if dependabot can handle kubernetes version updating (at least I have not seen it https://github.com/kubernetes-csi/external-provisioner/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+is%3Aclosed).

Oh.. yes, it can as long as its not excluded in package list.. we can also see such instance here . https://github.com/kubernetes-csi/volume-data-source-validator/pulls @msau42

humblec:
> In terms of whether we should wait for 1.26.0, I think it's fine to update to rc.0 now, and then update once more to 1.26.0 before we do the final release. That way we can start running tests with the new deps sooner. I don't expect signficant changes between rc.0 and .0

Ok.. if the plan is to get the tests run early,  it should be fine . :+1:

go.mod
```diff
@@ -58,10 +58,20 @@ require (
...
-	k8s.io/component-base v0.25.4 // indirect
+	k8s.io/component-base v0.26.0-rc.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20221110221610-a28e98eb7c70 // indirect
 	k8s.io/utils v0.0.0-20221108210102-8e77b1f39fe2 // indirect
 	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace k8s.io/api => k8s.io/api v0.26.0-rc.0
```
humblec, line 69: I think its better to remove this replace section.

xing-yang: +1

msau42: It looks like the script is relying on the replace statements to do special handling for the kubernetes libraries that don't follow the same naming scheme. @pohly do you know if it is safe to skip this logic if the repo does not import k/k?

https://github.com/kubernetes-csi/csi-release-tools/blob/master/go-get-kubernetes.sh#L125

pohly: If the repo doesn't use k/k, then it should be safe to drop the replace statements.

PR #395, opened 2022-12-12
Update dependency go modules for k8s v1.26.0

sunnylovestiramisu:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.26.0


jsafrane:
/release-note-none
?

jsafrane:
/lgtm
bad day today :-)

humblec:
@sunnylovestiramisu looks like some dependencies are still with rc tag https://github.com/kubernetes-csi/external-attacher/pull/395/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L17 or see #396 . considering you have filed PRs in sidecar repos, it would be better to revisit the existing PRs too

humblec:
> @humblec good catch, I missed that one.

yw :+1:

sunnylovestiramisu:
I was running the script go-get-kubernetes.sh and this is what the script generated. Do we usually manually update dependencies after running the script? It happened twice already for both 1.26.0-rc and 1.26.0.

msau42:
I think it's worth digging more into the script to figure out why it's not updating those. We want to get to a state where the script is reliable and doesn't require modifying by hand afterwards

PR #411, opened 2023-01-27
add option to limit grpc logging length

leiyiz:
**What type of PR is this?**
/kind feature

ListVolume call generates tons of log items because by default the entirety of the grpc response body gets logged out. That consumes log space really fast and we want to limit how much can be logged in the same message.

adding a flag to optionally limit GRPC log message length to fix this issue.

Fixes #

```release-note
Logging for gRPC call can now be limited configurably with flag `--grpc-log-char-cap`
```

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -76,6 +76,8 @@ var (
 
 	kubeAPIQPS   = flag.Float64("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
 	kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
+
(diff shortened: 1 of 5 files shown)


cmd/csi-attacher/main.go
```diff
@@ -76,6 +76,8 @@ var (
 
 	kubeAPIQPS   = flag.Float64("kube-api-qps", 5, "QPS to use while communicating with the kubernetes apiserver. Defaults to 5.0.")
 	kubeAPIBurst = flag.Int("kube-api-burst", 10, "Burst to use while communicating with the kubernetes apiserver. Defaults to 10.")
+
+	grpcLogCharCap = flag.Int("grpc-log-char-cap", -1, "The maximum amount of characters logged for every grpc responses. Defaults to no limit")
```
ggriffiths, line 80: Maybe name this `max-grpc-log-length`? This would match the grpc function more closely.

leiyiz: done

PR #495, opened 2023-10-03
Fixes CVE-2022-41723 

Sneha-at:
**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Bumps dependency to fix CVE-2022-41723

Fixes #

```release-note
NONE
```


go.mod
```diff
@@ -48,11 +48,11 @@ require (
 	github.com/prometheus/common v0.32.1 // indirect
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	golang.org/x/net v0.1.1-0.20221027164007-c63010009c80 // indirect
+	golang.org/x/net v0.8.0 // indirect
```
sunnylovestiramisu, line 51: The CVE only requires to update golang.org/x/net to >= v0.7.0. Is it required to update all the other things including k8s stuff?

Sneha-at: I updated the `client-go` package as it was dependent on this so `go mod tidy` updated all the packages. Should I update just `golang.org/x/net : 0.7.0`

Sneha-at: I updated the replace statement for x/net to point to `v0.8.0` version. Only removing the `replace` added x/net `v0.6.0` and `v0.0.0` dependencies in go.sum.

sunnylovestiramisu:
Testgrid is red before this PR submission: https://testgrid.k8s.io/sig-storage-csi-external-resizer#1-27-on-kubernetes-1-27

msau42:
Generally we submit updates to master branch first and then cherry pick to release branches to make sure we don't drop the fix in new versions. Can you also create a PR against master branch?

go.mod
```diff
@@ -69,4 +69,4 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace golang.org/x/net => golang.org/x/net v0.4.0
+replace golang.org/x/net => golang.org/x/net v0.8.0
```
msau42, line 72: Also in general, having replace statements is not going to work well with dependabot. Are we able to explicit set the version in the "require" section?

Sneha-at: This is an indirect dependency can we add it in require section? The other option I tried was to bump client-go to 1.24.17 as `golang/x/net` is it's dependency

Sneha-at:
In master branch the package is already [updated](https://github.com/kubernetes-csi/external-attacher/blob/8dac68df59d66a0f534336268c5197cc81886086/go.mod#L55) to `golang/x/net: 0.14.0`  so we should not have this CVE

sunnylovestiramisu:
Related to OSS issue: https://github.com/kubernetes/kubernetes/issues/120997

msau42:
> In master branch the package is already [updated](https://github.com/kubernetes-csi/external-attacher/blob/8dac68df59d66a0f534336268c5197cc81886086/go.mod#L55) to `golang/x/net: 0.14.0` so we should not have this CVE

It looks like 4.2 branch and below need to be updated: https://github.com/kubernetes-csi/external-attacher/blob/release-4.2/go.mod

go.mod
xing-yang: We can't move up to golang.org/x/net v0.17.0?

msau42:
3.5.x is out of support. It looks like this is addressed in the latest branches.

/close

PR #523, opened 2023-12-21
Update github.com/kubernetes-csi/csi-lib-utils to v0.16.0

msau42:
Ran ./go-modules-targeted-update.sh

```release-note
NONE
```


msau42:
still not right
/close

PR #561, opened 2024-05-29
Fix timeout on short CSI calls during attacher initialization

Fricounet:
**What type of PR is this?**

/kind bug

The same context.WithTimeout was used in different places when calling the CSI drivers in the main function. With recent changes, the code execution started taking longer and the context would hit its timeout on the `GetPluginCapabilities` driver call. For instance with when running with the azuredisk-csi-driver (I've seen this on the aws-ebs and the gcp-pd drivers too):
```
❯ k logs -c csi-attacher csi-azuredisk-controller-69fdb787c6-k59fw
I0529 12:26:30.367859       1 main.go:109] "Version" version="v4.6.0-0-gc58f4277"
I0529 12:26:30.369489       1 connection.go:234] "Connecting" address="unix:///csi/csi.sock"
I0529 12:26:31.370120       1 common.go:145] "Probing CSI driver for readiness"
I0529 12:26:31.370148       1 connection.go:264] "GRPC call" method="/csi.v1.Identity/Probe" request="{}"
I0529 12:26:31.371908       1 connection.go:270] "GRPC response" response="{\"ready\":{\"value\":true}}" err=null
I0529 12:26:31.371934       1 connection.go:264] "GRPC call" method="/csi.v1.Identity/GetPluginInfo" request="{}"
I0529 12:26:31.372446       1 connection.go:270] "GRPC response" response="{\"name\":\"disk.csi.azure.com\",\"vendor_version\":\"v1.30.1\"}" err=null
I0529 12:26:31.372466       1 main.go:169] "CSI driver name" driver="disk.csi.azure.com"
I0529 12:26:31.373062       1 connection.go:234] "Connecting" address="unix:///csi/csi.sock"
I0529 12:26:32.374364       1 common.go:145] "Probing CSI driver for readiness"
I0529 12:26:32.374396       1 connection.go:264] "GRPC call" method="/csi.v1.Identity/Probe" request="{}"
I0529 12:26:32.374869       1 connection.go:270] "GRPC response" response="{\"ready\":{\"value\":true}}" err=null
I0529 12:26:32.374921       1 connection.go:264] "GRPC call" driver="disk.csi.azure.com" method="/csi.v1.Identity/GetPluginCapabilities" request="{}"
I0529 12:26:32.375015       1 main.go:196] "ServeMux listening" driver="disk.csi.azure.com" address=":22012" metricsPath="/metrics"
I0529 12:26:32.375048       1 connection.go:270] "GRPC response" driver="disk.csi.azure.com" response="{}" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded"
E0529 12:26:32.375069       1 main.go:207] "Failed to check if the CSI Driver supports the CONTROLLER_SERVICE" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded" driver="disk.csi.azure.com"
```

 This change uses a separate context.WithTimeout for each short CSI call using the default `csiTimeout` defined in main.

Fixes #

```release-note
Fixed an issue where the attacher would see its context timeout on startup when calling the CSI driver.
```

### DIFF
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -167,7 +167,6 @@ func main() {
 	}
 	logger = klog.LoggerWithValues(logger, "driver", csiAttacher)
 	logger.V(2).Info("CSI driver name")
-	cancelationCtx = klog.NewContext(cancelationCtx, logger)
 
 	translator := csitrans.New()
 	if translator.IsMigratedCSIDriverByName(csiAttacher) {
@@ -202,6 +201,9 @@ func main() {
 		}()
 	}
 
+	cancelationCtx, cancel = context.WithTimeout(ctx, csiTimeout)
+	cancelationCtx = klog.NewContext(cancelationCtx, logger)
+	defer cancel()
 	supportsService, err := supportsPluginControllerService(cancelationCtx, csiConn)
 	if err != nil {
 		logger.Error(err, "Failed to check if the CSI Driver supports the CONTROLLER_SERVICE")


huww98:
`connection.go:234] "Connecting" address="unix:///csi/csi.sock"` this seems taking exactly 1 second from your logs. Do you know why? It looks like the CSI response is very fast, and 1 second timeout should be enough for both RPC calls.

Fricounet:
@huww98 what you indicate is a log that does not come from the same call to the CSI driver. It comes from the [Connect](https://github.com/kubernetes-csi/csi-lib-utils/blob/master/connection/connection.go#L234) method of CSI lib utils which is triggered [here](https://github.com/kubernetes-csi/external-attacher/blob/eeed54970cdaee747be9cee3f9f496cc9d26a7ed/cmd/csi-attacher/main.go#L174) in the code. It looks like the timeout appears 1s after this because the context is configure in the call to get the driver name just [before](https://github.com/kubernetes-csi/external-attacher/blob/eeed54970cdaee747be9cee3f9f496cc9d26a7ed/cmd/csi-attacher/main.go#L163). You can even see this call in the logs `connection.go:264] "GRPC call" method="/csi.v1.Identity/GetPluginInfo" request="{}"`

While in the bug, the timeout appears in a later call to the csi driver [here](https://github.com/kubernetes-csi/external-attacher/blob/eeed54970cdaee747be9cee3f9f496cc9d26a7ed/cmd/csi-attacher/main.go#L209)

Fricounet:
I've figured where this delay comes from and It is not an issue with the connection to the driver but with a change to the `Probe` function that is run just after. In the [latest version](https://github.com/kubernetes-csi/csi-lib-utils/commit/ca88d98a5cae3e150849392a99387c6e27e6bebb#diff-5fb7c6e20b635d2fb705b5f7ec97d8f1a35c802d9bd76105b814b43ee4885c6eR138) of csi-lib-utils, it was changed to use a `Ticker` instead of the sleep it was using before between each probe. As a result, the first probe is only launched after 1s instead of immediately resulting in the delay we see

Fricounet:
I've submitted a PR in https://github.com/kubernetes-csi/csi-lib-utils/pull/175 to restore the previous behavior of the Probe. But this PR stays relevant IMHO because the timeout should be used for [short csi calls](https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L51) and having it shared for 2 calls separated by potentially long running calls (Connection and Probe for instance) feels wrong.

jsafrane:
/lgtm
/approve

You're right both about the ProbeForever being slower and that the timeout context should start just before corresponding call.

Fricounet:
And thanks for all our work on CSI and storage!

Issue #609, opened 2024-12-13
Release lease on SIGTERM


jsafrane:
When the sidecar is terminating because of SIGTERM, it should finish all running operations + release its leader election lease. This will allow another attacher sidecar to pick the lease quickly, without waiting for leader election timeouts.

"Finish all running operations" could be accomplished by cancelling a global `context.Context` +waiting for workqueues to finish.

PR #633, opened 2025-03-06
Restart external-attacher faster by releasing leader election lease on sigterm

rhrmo:
**What type of PR is this?**
/kind bug

This PR makes external-attacher restart faster when it is terminating because of sigterm. This is accomplished by releasing leader election lease on sigterm and then it does not have to wait for lease to expire but it proceeds with restart right away. This is the first PR of series of PRs that fix https://github.com/kubernetes-csi/external-attacher/issues/609

Fixes #

This PR is the first of the series of PRs that fix https://github.com/kubernetes-csi/external-attacher/issues/609 and I want to use it to agree on a solution - if you have any suggestions, or find anything wrong, please add a comment and let me know.
I had to import newer version of csi-lib-utils library, so that it would contain changes from https://github.com/kubernetes-csi/csi-lib-utils/pull/191 (go get github.com/kubernetes-csi/csi-lib-utils@f77445f75e9c4954062cdddcef4863adab21947b), so before merging this we would also need to have new version of csi-lib-utils released.

```release-note
NONE
```


jsafrane:
I tested in on a cluster.

What's great is that when the attacher gets SIGTERM, it exits immediately (this PR does not change that) + release its lease, so a newly started attacher starts immediately. This is taken when SIGTERM was received in the middle of ControllerPublish:

```
I0310 15:57:31.102801  133591 connection.go:264] "GRPC call" VolumeAttachment="csi-043731379aa11361179f9e22a17f1fda8d2b0716696126267433bf49892e92b2" method="/csi.v1.Controller/ControllerPublishVolume" request="{\"node_id\":\"192.168.124.90\",\"volume_capability\":{\"access_mode\":{\"mode\":\"SINGLE_NODE_MULTI_WRITER\"},\"mount\":{}},\"volume_context\":{\"storage.kubernetes.io/csiProvisionerIdentity\":\"1741618771172-2394-hostpath.csi.k8s.io\"},\"volume_id\":\"5a4de638-fdc0-11ef-b375-5ed7a0a04918\"}"

[waiting for ControllerPublishVolumeResponse]
^C

I0310 15:57:32.373544  133591 main.go:292] Received SIGTERM or SIGINT signal, shutting down controller.
I0310 15:57:32.373594  133591 controller.go:151] "Shutting CSI attacher"
I0310 15:57:32.373682  133591 reflector.go:319] Stopping reflector *v1.CSINode (10m0s) from k8s.io/client-go/informers/factory.go:160
I0310 15:57:32.373719  133591 reflector.go:319] Stopping reflector *v1.PersistentVolume (10m0s) from k8s.io/client-go/informers/factory.go:160
I0310 15:57:32.373748  133591 reflector.go:319] Stopping reflector *v1.VolumeAttachment (10m0s) from k8s.io/client-go/informers/factory.go:160
I0310 15:57:32.373719  133591 connection.go:270] "GRPC response" VolumeAttachment="csi-043731379aa11361179f9e22a17f1fda8d2b0716696126267433bf49892e92b2" response="{}" err="rpc error: code = Canceled desc = context canceled"
I0310 15:57:32.373761  133591 csi_handler.go:611] "Saving attach error" VolumeAttachment="csi-043731379aa11361179f9e22a17f1fda8d2b0716696126267433bf49892e92b2"
I0310 15:57:32.374421  133591 csi_handler.go:267] "Failed to save attach error to VolumeAttachment" VolumeAttachment="csi-043731379aa11361179f9e22a17f1fda8d2b0716696126267433bf49892e92b2" err="client rate limiter Wait returned an error: context canceled"
I0310 15:57:32.374448  133591 csi_handler.go:243] "Error processing" VolumeAttachment="csi-043731379aa11361179f9e22a17f1fda8d2b0716696126267433bf49892e92b2" err="failed to attach: rpc error: code = Canceled desc = context canceled"
E0310 15:57:32.376055  133591 event.go:464] "Unable to record event (will not retry!)" err="broadcaster already stopped"
E0310 15:57:32.376066  133591 leader_election.go:198] "Stopped leading"
```

What's not so great is that the CSI driver does not know about any of this. It got `ControllerPublishVolume`, it is still attaching the volume and eventually it may succeed. But at the same time, another external-attacher might get the lease very quickly and call the same `ControllerPublishVolume` in parallel.

This could happen also today, however, before this PR, the new attacher would need to wait for the old attacher lease to expire (15 seconds in the default config). Which could give the old CSI driver enough time either complete the attachment or die (because it's unlikely that the attacher got SIGTERM and the CSI driver in the same Pod did not).

cmd/csi-attacher/main.go
```diff
@@ -279,6 +280,18 @@ func main() {
 		*reconcileSync,
 	)
 
+	// handle SIGTERM and SIGINT by cancelling the context.
+	shutdownCtx, cancel := context.WithCancel(ctx)
+	shutdownHandler := server.SetupSignalHandler()
+	ctx, terminate := context.WithCancel(shutdownCtx)
```
gnufied, line 286: That ctx handling looks weird. We wrap existing `ctx` with `WithCancel` and then wrap it again with - `WithCancel`. Is that expected?

rhrmo: I cannot think of an answer from the top of my head, but I think that is a remainder I forgot to delete, I will double check and edit the code or give an explanation why we should use it.

msau42:
> What's not so great is that the CSI driver does not know about any of this. It got `ControllerPublishVolume`, it is still attaching the volume and eventually it may succeed. But at the same time, another external-attacher might get the lease very quickly and call the same `ControllerPublishVolume` in parallel.

Instead of shutting down immediately on a sigterm, can we wait until all pending requests are finished? It would probably require changes to our reconciler.

jsafrane:
>  Instead of shutting down immediately on a sigterm, can we wait until all ongoing requests are finished? It would probably require changes to our reconciler.

We would need two contexts then and it might get messy which context should be used when.

@rhrmo can you explore this idea? I think the only place that should unblock when a signal is received would be here (to call `ShutDown` on all workqueues, which will then wait for existing work to finish):
https://github.com/kubernetes-csi/external-attacher/blob/fbf2dba68fc0c1aa49c787a021e28490e9230154/pkg/controller/controller.go#L150

And here to finish periodic reconcile:
https://github.com/kubernetes-csi/external-attacher/blob/fbf2dba68fc0c1aa49c787a021e28490e9230154/pkg/controller/controller.go#L142

Maybe you find other places. But it will be different in each sidecar.

rhrmo:
Sure thing!

cmd/csi-attacher/main.go
```diff
@@ -279,6 +280,18 @@ func main() {
...
+	defer terminate()
+
+	go func() {
+		defer cancel()
+		<-shutdownHandler
```
sunnylovestiramisu, line 291: Does this work on Windows nodes too? SIGTERM behaves a bit different in Windows than Linux?

rhrmo: Good cache! I think on windows it does os.interupt ([here](https://github.com/kubernetes/kubernetes/blob/34e80be1330556848521314ece92e9277efb26d4/staging/src/k8s.io/apiserver/pkg/server/signal_windows.go#L23) is how kubernetes handles it AFAIK) and just kills the process, I am not sure if we can cache this signal... I will test this with the updated code (I should update the PR by Wednesday based on previous comments). If I cannot think of a way to make this work for Windows we can discuss how we want to handle it.

rhrmo: sorry for the delay, I have some issues with testing the code and cluster creation... I will update once I handle these issues 😅

jsafrane: @sunnylovestiramisu Kubernetes does not ship any Windows containers with the CSI sidecars. We build just the Linux containers ([here](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-sig-storage.yaml))

Vendors, e.g. Microsoft, ships the driver registrar and liveness probe (see e.g. [here](https://github.com/kubernetes-csi/csi-driver-smb/blob/475af8ac63f67cc0e40daba5a0d07717e92e1148/deploy/csi-smb-node-windows-hostprocess.yaml#L45C18-L45C88)), but even them don't seem to ship the Windows version of the attacher, provisioner, resizer or snapshotter.

Do we really want to support these sidecars on Windows? Nobody has probably run them there.

jsafrane: I asked on [slack](https://kubernetes.slack.com/archives/C0SJ4AFB7/p1744723105562959), even Microsoft runs the sidecars on Linux

jsafrane:
>> What's not so great is that the CSI driver does not know about any of this. It got ControllerPublishVolume, it is still attaching the volume and eventually it may succeed. But at the same time, another external-attacher might get the lease very quickly and call the same ControllerPublishVolume in parallel.

> Instead of shutting down immediately on a sigterm, can we wait until all ongoing requests are finished? It would probably require changes to our reconciler.

I can see there is a code that waits for workers to finish before exiting, still, they should finish with a lot of `context canceled` errors, because it's the main context that gets cancelled.

cmd/csi-attacher/main.go
```diff
@@ -280,11 +296,37 @@ func main() {
...
+	defer terminate()
+
+	var signalReceived bool
+
+	go func() {
+		defer cancelRun()
+		<-shutdownHandler
+		signalReceived = true
+		logger.Info("Received SIGTERM or SIGINT signal, shutting down controller.")
+	}()
```
jsafrane, line 313: All this should be behind the feature gate too. We don't want to change the existing behavior.

pkg/controller/controller.go
```diff
@@ -133,18 +136,46 @@ func (ctrl *CSIAttachController) Run(ctx context.Context, workers int) {
...
-			err := ctrl.handler.ReconcileVA(ctx)
-			if err != nil {
-				logger.Error(err, "Failed to reconcile VolumeAttachment")
-			}
-		}, ctrl.reconcileSync)
+		if ctrl.shouldReconcileVolumeAttachment {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				go wait.UntilWithContext(ctx, func(ctx context.Context) {
```
jsafrane, line 157:
```suggestion
				wait.UntilWithContext(ctx, func(ctx context.Context) {
```

Otherwise the outer `go func()` finishes instantly.

cmd/csi-attacher/main.go
```diff
@@ -313,6 +364,10 @@ func main() {
 		le.WithLeaseDuration(*leaderElectionLeaseDuration)
 		le.WithRenewDeadline(*leaderElectionRenewDeadline)
 		le.WithRetryPeriod(*leaderElectionRetryPeriod)
+		if utilfeature.DefaultFeatureGate.Enabled(features.ReleaseLeaderElectionOnExit) {
+			le.WithReleaseOnCancel(true)
+			le.WithContext(ctx)
```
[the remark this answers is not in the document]
rhrmo: Perhaps it would make more sense to exchange the names of ctx and globalCtx. Currently when globalCtx gets shutdown, the whole process gets shutdown, but it is derived from ctx, so the name change could clarify the contexts use. 
Another alternative I can think of is renaming globalCtx to shutdownCtx or lifecycleCtx or something along those lines, but I'm not sure if it's fitting.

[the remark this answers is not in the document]
jsafrane: If I understand it correctly, `ctx` becomes basically unused when `ReleaseLeaderElectionOnExit` is set. We could remove `globalCtx` and use `ctx, terminate = context.WithCancel(ctx)`. One less context to worry about.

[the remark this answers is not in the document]
rhrmo: You're correct, I guess we don't need to differentiate them then and just use ctx. I will update it.

cmd/csi-attacher/main.go
```diff
@@ -280,11 +296,45 @@ func main() {
...
 	run := func(ctx context.Context) {
-		stopCh := ctx.Done()
-		factory.Start(stopCh)
-		ctrl.Run(ctx, int(*workerThreads))
+		if utilfeature.DefaultFeatureGate.Enabled(features.ReleaseLeaderElectionOnExit) {
+			var wg sync.WaitGroup
+			stopCh := shutdownHandler
+			factory.Start(stopCh)
+			ctrl.Run(controllerCtx, int(*workerThreads), &wg)
+			if signalReceived {
```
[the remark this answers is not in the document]
rhrmo: I removed the whole `signalReceived` variable, testing `Run()` is bit tricky though because it waits for the context to be cancelled before it returns, by trying to call `return` at it's end instead `<-ctx.Done()` it finished without any issues.

PR #620, opened 2025-01-13
[draft][DNM]Refactor command parameters initialize

mowangdk:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
/kind cleanup

**What this PR does / why we need it**:
Refactor command parameters initialize for csi sidecar aios
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #652, opened 2025-05-31
Separate attacher flags into attacher only flags and global shared flags

mauriciopoppe:
**What type of PR is this?**
>
/kind design

Split attacher flags into attacher only flags and global flags.

We're moving some of the flags into its own package so that the flags can be imported by `external-attacher` and by other repositories. We're doing this to prepare external-attacher for its integration with the CSI Sidecar monorepo.

Testing:

- Within this repo there's no op besides checking that it builds fine.
- In the CSI Sidecar monorepo we should be able to import the new `cmd/config` package and reuse the exported functions and types. The CSI Sidecar monorepo will use the function `RegisterAttacherFlagsWithPrefix` so that the attacher flags are imported with the `attacher-` prefix.

In the CSI Sidecar monorepo the change to adopt this is https://github.com/mauriciopoppe/csi-sidecars-aio-poc/pull/11. 

After compiling the CSI Sidecar monorepo binary the output for its help is:

```
+ ./bin/csi-sidecars --help
Usage of ./bin/csi-sidecars:
      --add_dir_header                                    If true, adds the file directory to the header of the log messages
      --alsologtostderr                                   log to standard error as well as files (no effect when -logtostderr=true)
      --attacher-default-fstype string                    The default filesystem type of the volume to use.
      --attacher-max-entries int                          Max entries per each page in volume lister call, 0 means no limit.
      --attacher-max-grpc-log-length int                  The maximum amount of characters logged for every grpc responses. Defaults to no limit (default -1)
      --attacher-reconcile-sync duration                  Resync interval of the VolumeAttachment reconciler. (default 1m0s)
      --attacher-retry-interval-max duration              Maximum retry interval of failed create volume or deletion. (default 5m0s)
      --attacher-retry-interval-start duration            Initial retry interval of failed create volume or deletion. It doubles with each failure, up to retry-interval-max. (default 1s)
      --attacher-timeout duration                         Timeout for waiting for attaching or detaching the volume. (default 15s)
      --attacher-worker-threads int                       Number of worker threads per sidecar (default 10)
```

For more info please read https://docs.google.com/document/d/1AKqJeAlBL8PkH8D9zABCZ82Bk1N46EygKPvVh5p4-qU/edit?tab=t.0.

```release-note
NONE
```

/cc @jsafrane @xing-yang @mowangdk @ConnorJC3

### DIFF
--- a/cmd/csi-attacher/config/flags.go
+++ b/cmd/csi-attacher/config/flags.go
@@ -1,0 +1,38 @@
+package config
+
+import (
+	"flag"
+	"time"
+)
+
+type AttacherConfiguration struct {
+	MaxEntries         int
+	ReconcileSync      time.Duration
+	MaxGRPCLogLength   int
+	WorkerThreads      int
+	DefaultFSType      string
+	Timeout            time.Duration
+	RetryIntervalStart time.Duration
+	RetryIntervalMax   time.Duration
+}
+
+func registerAttacherFlags(flags *flag.FlagSet, configuration *AttacherConfiguration, prefix string) {
+	flag.IntVar(&configuration.MaxEntries, prefix+"max-entries", 0, "Max entries per each page in volume lister call, 0 means no limit.")
+	flag.DurationVar(&configuration.ReconcileSync, prefix+"reconcile-sync", 1*time.Minute, "Resync interval of the VolumeAttachment reconciler.")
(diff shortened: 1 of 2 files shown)


mowangdk:
So is this just a demo PR for the csi aio integration or will it be a former PR in the future? I think maybe I can close #620 and  we can refine this PR. they are basically the same

PR #662, opened 2025-07-07
Populate `VolumeError.ErrorCode` field in VolumeAttachment object

torredil:
**What type of PR is this?**

/kind api-change

[KEP-4876: Mutable CSINode Allocatable Property](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/4876-mutable-csinode-allocatable) introduces a new field in the `VolumeError` struct used in the `VolumeAttachment` object Status field:

```
// Captures an error encountered during a volume operation.
type VolumeError struct {
  // errorCode is a numeric gRPC code representing the error encountered during Attach or Detach operations.
    ErrorCode *int32
}
```

This change populates the field when a plugin's ControllerPublishVolume RPC returns an error.

N/A

See API strategy for the VolumeAttachment object here:

https://github.com/kubernetes/kubernetes/blob/ee012e883f863a141cab35e5a76cbe5d378ad36f/pkg/registry/storage/volumeattachment/strategy.go#L138

```release-note
Populate VolumeError.ErrorCode field in VolumeAttachment object.
```


pkg/controller/csi_handler.go
```diff
@@ -610,11 +613,21 @@ func (h *csiHandler) saveAttachError(ctx context.Context, va *storage.VolumeAtta
...
 	logger.V(4).Info("Saving attach error")
 	clone := va.DeepCopy()
-	clone.Status.AttachError = &storage.VolumeError{
+
+	volumeError := &storage.VolumeError{
 		Message: err.Error(),
 		Time:    metav1.Now(),
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.MutableCSINodeAllocatableCount) {
```
gnufied, line 622: IMO, you will have to first define external-attacher only feature gates. You can't read feature gates from k8s this way.

See - https://github.com/kubernetes-csi/external-attacher/blob/master/cmd/csi-attacher/main.go#L98 and how other sidecars define sidecar local feature gates.

torredil: Thank you Hemant 💯 , published new rev doing the needful.

PR #733, opened 2026-06-19
Log work queue backlog to surface queued attach/detach work

### LINKED ISSUE #706: Invisible backlog of queued work
## Problem

The external-attacher maintains an invisible backlog of work which is not revealed in its logs or in an emitted metric

## So what

* Service operators do not have any indication that they have misconfigured the external-attacher
* The backlogs result in an inability to complete the volume attach/detach lifecycle within a reasonable amount of time
* Stateful services performing deployments experience reduced availability for extended periods of time

## Proposal

Emit metrics and/or logs about the backlog and progress against the backlog by default

## What happened

My service began to observe some pods taking 45 minutes to come back to active despite being successfully rescheduled.  Everything appeared fine except that the volumes weren't attached.  At first it was just one or two, and then it was hundreds.   Here's a timeline for an example where a detached volume was not attached to the new node for 27 minutes:

```
06:36:22 — Old node unmounts volume
06:36:24 — kube-controller-manager started detach from old node
06:38:24 — KCM started attach attempt on new node indicating the pod had been successfully rescheduled (volume still attached to old node)
06:44:42 — Last log with "volume attachment is being deleted" in KCM
06:45:44 — CSI controller - logged detaching
06:45:59 — CSI controller - detach completed from old node
< 27 minute gap >
07:13:30 — CSI controller attached volume to new node
07:13:32 — CSI controller attach completed
```

You'll notice that there aren't any logs in this timeline about the external-attacher because by default nothing was emitting that indicating it was doing anything (or waiting).  It took the help of EKS, EBS, and EC2 to pinpoint the problem because the log entries across all involved components were essentially barren with respect to the volume-ids, pvcs, or the pod.  Nothing was happening and nobody knew why.

The fix ended up being that I needed to increase the `worker-threads` configuration from 10 to 1,000 in order to remove the bottleneck from external-attacher.  It had built up a queue of thousands of volumes and it wasn't signaling that it was underwater.  This configuration is [well documented](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/faq.md#driver-performance-for-large-scale-clusters) in the ebs-csi-driver readme for large-scale clusters, but it requires you to be psychically aware that the problem exists in the external-attacher in the first place, or what large-scale means.  The default workers has recently been increased to 100, which will be plenty for most service owners, but my problem didn't go away even after I'd increased from 10 to 300.  Service owners at every scale need a way to help calibrate the external-attacher's throughput.

kishiel:
The external-attacher processes VolumeAttachment and PersistentVolume events through rate-limited work queues. When the queues build up a backlog (for example, when --worker-threads is set too low for the attach/detach load), there is no signal in the logs. Operators only see individual items being processed, not the size of the pile waiting behind them, which makes it hard to tell that the attacher is overwhelmed and that throughput needs to be increased.

Add a periodic log line that reports the depth of the VolumeAttachment and PersistentVolume queues. It is emitted only while a backlog exists, so a healthy attacher with empty queues stays quiet. The interval is controlled by the new --log-backlog-interval flag (default 30s, 0 to disable). The depth is also already exported as the workqueue_depth metric; this makes the same signal visible to anyone reading logs without a metrics pipeline.

**What type of PR is this?**
/kind feature

Adds logging visibility to the queue depth for the controller which will clue an operator into a developing backlog of work.  Without this change operators do not have log events which will lead them to conclude they do not have enough worker threads.

Fixes #706 

Good luck everyone, this is my first `go` change

```release-note
Output log events when a backlog of queued work exists
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Note that the external-attacher does not scale with more replicas. Only one exte
 
 * `--reconcile-sync`: Resync frequency of the attached volumes with the driver. See [Periodic re-sync](#periodic-re-sync) for details. 1 minute is used by default.
 
+* `--log-backlog-interval`: Interval at which the depth of the VolumeAttachment and PersistentVolume work queues is logged whenever there is a backlog. A growing backlog usually means `--worker-threads` is too low for the current attach/detach load. The depth is also exported as the `workqueue_depth` metric. 30 seconds is used by default. Set to 0 to disable.
+
 * `--kube-api-qps`: The number of requests per second sent by a Kubernetes client to the Kubernetes API server. Defaults to `5.0`.
 
 * `--kube-api-burst`: The number of requests to the Kubernetes API server, exceeding the QPS, that can be sent at any given time. Defaults to `10`.
--- a/cmd/csi-attacher/main.go
+++ b/cmd/csi-attacher/main.go
@@ -75,6 +75,8 @@ var (
 
 	reconcileSync = flag.Duration("reconcile-sync", 1*time.Minute, "Resync interval of the VolumeAttachment reconciler.")
 
+	logBacklogInterval = flag.Duration("log-backlog-interval", 30*time.Second, "Interval at which the depth of the VolumeAttachment and PersistentVolume work queues is logged when there is a backlog. Set to 0 to disable.")
+
 	maxGRPCLogLength = flag.Int("max-grpc-log-length", -1, "The maximum amount of characters logged for every grpc responses. Defaults to no limit")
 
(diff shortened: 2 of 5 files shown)


jsafrane:
I think a metric would be much better for this than spamming logs, potentially with an alert that would notify when the queue is too long. 

In addition, all CSI sidecars will have the same problem. It belongs to csi-lib-utils + all sidecars. Prototyping here is fine, to see the whole picture in a single repo, but eventually I'll ask to move the code that can be shared + update all sidecars that have queues.

